PR #46, opened 2015-01-20
Add rerun feature

nakaji-s:
Rerun feature:  
bin/rerun runs only tests failed in the previous run of bin/test.

Motivation:  
We sometimes want to run only failed tests. But using `-focus` is a bit painful.  
ex) `$ bin/test -focus="failedtest1 | failedtest2 | failedtest3..."`

Now we can rerun failed tests just with the number of retries.  
ex) `$ bin/rerun 1`

### DIFF
--- a/apps/init_test.go
+++ b/apps/init_test.go
@@ -39,11 +39,15 @@ func TestApplications(t *testing.T) {
 	environment := helpers.NewEnvironment(context)
 
 	BeforeSuite(func() {
-		environment.Setup()
+		InterceptGomegaFailures(func() {
+			environment.Setup()
+		})
 	})
 
 	AfterSuite(func() {
-		environment.Teardown()
+		InterceptGomegaFailures(func() {
+			environment.Teardown()
+		})
 	})
 
 	componentName := "Applications"
--- a/bin/rerun
+++ b/bin/rerun
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+if ! which go 2>/dev/null; then
+  echo "Go is not installed."
+  exit 1
+fi
+
+if [ -z "$GOPATH" ]; then
+  echo "\$GOPATH is not set."
+  exit 1
+fi
+
+if [ $# -ne 1 ]; then
+  echo "need to set an arg(max repeat count)" 1>&2
+  exit 1
+fi
+
+. $(dirname $0)/goenv
+set +ex
+
+num=1
+while [ $num -le $1 ]
+do
+  echo "/##############################################################/"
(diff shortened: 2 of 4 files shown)


dieucao:
@nakaji-s Are you covered by a CLA from your organization? Could you make your membership of nttlabs public?

nakaji-s:
I'm sorry about that. I changed my membership of nttlabs to public.

luan:
Hi @nakaji-s,

You mentioned that the motivation for this is that sometimes you want to rerun only failed tests. Can you elaborate a little more on that? At first sight, it seems like this is hiding a bigger issue: tests are flaky.

We'd like our tests not to be flaky rather than working around the flakiness by re-running them.
If you have any other motive for this we could be convinced, otherwise we'd rather understand why they fail for you and try to work on fixing that.

Thanks,
@luan & @gtona, CF Runtime Team

nakaji-s:
Thanks for reading my pull request and comments.

I agree with you that tests shouldn't be "flaky."
 
But tests could be flaky because they depends on the situation, especially the environment they are running on.Our environment and your environment is different, and another one's environment may also be different.And there are many timeout settings and timing-dependent tests (ex: cli should have received "THIS MESSAGE" but at that time the request may have yet processed), so I think it is not possible to eliminate flakiness at all.

With the "rerun" feature, at least it makes easier for us to distinguish real problems from flakiness.

Additionally, If tests passed while bin/rerun, we can know the tests can be passed and we can choose next deal.(ex: set timeout params more longer). So the rerun feature can contribute to make tests less flaky.

Moreover, it improves each test operator's peace of mind. If we don't have rerun, a test operator should wait until the CATs finishes, because she/he should rerun the tests manually if the first trial fails, to distinguish the cause is flakiness or not. With the rerun feature, the operator just starts the CATs and may go home to sleep. If the trial fails in the next morning, she/he is not to blame. It's a real problem and the developer to blame. :stuck_out_tongue_winking_eye:

Again, I agree with you that the rerun feature should not be used to ignore flakiness of the tests. And I think it can be used to make the tests less flaky.
 
Thanks.

morika-t:
Hi CF Runtime Team

Have you decided what to do with this PullRequest?

jfmyers9:
Hi @nakaji-s and @morika-t,

At this time, we do not want this feature in the CATS. We are making concerted efforts to reducing the flakeyness of the CATS and hopefully your issues will be resolved through improvements in tests. 

Thanks,

@jfmyers9, CF Runtime Team

PR #48, opened 2015-02-27
Fix unintentional overwrite of BROKER_START_TIMEOUT

nota-ja:
When the CATs config file (eg. integration_config.json)

- have `cf_push_timeout` and
- does not have `broker_start_timeout`

`BROKER_START_TIMEOUT` will be set to _ZERO_.
This seems an unintentional behavior.


crhino:
Thanks for the catch, @nota-ja!

@zaksoup & @crhino, CF Runtime Team

nota-ja:
Thanks for merging, @zaksoup & @crhino !

PR #49, opened 2015-03-12
Fix python fixture to work with both Python 2 and 3

styeung:
Existing fixture was breaking on Python versions 3+.

### DIFF
--- a/assets/python/server.py
+++ b/assets/python/server.py
@@ -1,5 +1,9 @@
 import sys
-from BaseHTTPServer import BaseHTTPRequestHandler,HTTPServer
+
+if int(sys.version[0]) >= 3:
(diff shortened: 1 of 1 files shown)


flavorjones:
Please note that Sai To is a Pivotal employee and is covered under the corporate CLA.

I've given him instructions on how to satisfy dreddbot's conditions by making his org membership public.

zaksoup:
Hi @styeung,

We decided to go with setting the python version in the `runtime.txt` file as we would like to avoid increasing complexity in these sample applications. See c5a30e0fbc4b0a81192e9620f2911cc6cd4b5a2d.

Thanks for bringing this to our attention,

@jfmyers9 && @zaksoup, CF Runtime Team

PR #51, opened 2015-03-30
Added a new test case for scaling of an App

shashidharatd:
CATS does not have a test case for testing app scaling feature whereas this test case is present in DATS, so adding this app scaling test case from DATS.

### DIFF
--- a/apps/instance_reporting_test.go
+++ b/apps/instance_reporting_test.go
@@ -1,0 +1,45 @@
+package apps
+
+import (
+	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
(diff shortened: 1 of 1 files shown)


utako:
@shashidharatd We see that the CATs have this test [here](https://github.com/cloudfoundry/cf-acceptance-tests/blob/master/apps/dynamic_info_test.go#L56) that seems to be doing a similar flow. Does that test cover the situation you are thinking of?

@crhino && @utako, CF Runtime Team

shashidharatd:
hi @utako,
  sorry, i did not notice this testcase. it covers the scenario which we want to test.

Thanks
@shashidharatd

PR #54, opened 2015-04-15
Adding Test Case for memory quota management

asifhuawei:
As quota management is a very important feature of CF so there should be a test case in CATS to validate this feature.

If a user pushes an app which needs a memory quota more than that of organization memory quota then it should fail while starting the app.

A similar type of test case for this is already there in [diego-acceptance-tests](https://github.com/cloudfoundry-incubator/diego-acceptance-tests/blob/master/diego/staging_failure_test.go#L45)

### DIFF
--- a/apps/staging_log_test.go
+++ b/apps/staging_log_test.go
@@ -5,10 +5,13 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gexec"
 
 	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
 	"github.com/cloudfoundry-incubator/cf-test-helpers/generator"
+	"github.com/cloudfoundry-incubator/cf-test-helpers/helpers"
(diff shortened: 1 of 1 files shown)


zaksoup:
Hi @asifhuawei,

Thanks for your pull request! We understand that quotas are incredibly important to our open-source partners and contributors. However, this functionality is [already tested](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/spec/unit/controllers/runtime/apps_controller_spec.rb#L645-L653) at the controller test level.

Zachary
CF Runtime Team

maximilien:
Also, to chime in as FYI @asifhuawei is that CATS need to be fast and generally should and have only testes happy paths and not all possible permutations of the "things" under test.

Just as a point of reference, current CATS, last time (two weeks ago) I ran them on a pretty powerful machine took more than 15 minutes to run. This is generally getting worst as new functionality that really need to be tested is added.

Hope this helps.

asifhuawei:
Hi @zaksoup @maximilien thanks for your input, I got your point.

PR #55, opened 2015-05-15
Assert correct staging error message by cli version

simonleung8:
CLI is now hitting /apps instead of /instances when polling for staging status, the error messages from /instances and /apps are different, hence we are updating CATs.

Included 2 changes

- new package for checking cli version
- assert the correct staging failing message depending on CLI version


utako:
Hi @simonleung8,

We feel  like this acceptance test will be more easily maintainable if we take out the version conditionals and just fail if the CLI version is not at or above the minimum version. That way, we can fail at a single point instead of having individual tests be dependent on a CLI version.

If you're able to update these tests, we'll be happy to merge your PR.

Thanks,
@wendorf and @utako, CF Runtime Team

dieucao:
@simonleung8 Closing this PR.  Please make a new PR with requested changes.

PR #61, opened 2015-07-16
improve cli version checking method

simonleung8:
Before, we are comparing version strings, this creates problem if we have version comparing such as "6.12.0">"6.9.0" (returns false), now checking is done by comparing major,minor,patch independently

Signed-off-by: Jonathan Berkhahn <jaberkha@us.ibm.com>


helpers/cli_version_check/cli_version/installed_cli_version.go
```diff
@@ -0,0 +1,42 @@
...
+
+type cliVersion struct {
+	fullVersionStr string
+}
+
+func NewInstalledCliVersion() InstalledCliVersion {
+	output, err := exec.Command("cf", "-v").Output()
+	if err != nil {
+		fmt.Println("Error trying to determine CF CLI version:", err)
+		os.Exit(1)
```
Amit-PivotalLabs: Rather than exiting, this should return an error that the upstream test asserts should not have occurred.

helpers/cli_version_check/cli_version/installed_cli_version.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	return &cliVersion{string(output)}
+}
+
+func (c *cliVersion) GetVersion() string {
+	re := regexp.MustCompile(`[0-9]+\.[0-9]+\.[0-9]+(?:\.[0-9]+)?`)
```
Amit-PivotalLabs: This should be an unexported package variable, not a local variable evaluated each time the method is called.

helpers/cli_version_check/cli_version_check_test.go
```diff
@@ -2,7 +2,7 @@ package cli_version_check_test
 
 import (
 	. "github.com/cloudfoundry/cf-acceptance-tests/helpers/cli_version_check"
-	"github.com/cloudfoundry/cf-acceptance-tests/helpers/cli_version_check/cli_version/fakes"
+	// "github.com/cloudfoundry/cf-acceptance-tests/helpers/cli_version_check/cli_version/fakes"
```
Amit-PivotalLabs: Please don't check in commented out imports.

helpers/cli_version_check/cli_version/installed_cli_version_test.go
```diff
@@ -0,0 +1,34 @@
...
+	Context("GetVersion", func() {
+		It("returns a version value that only contains numbers and dots", func() {
+			ver.SetFullVersionString("abcdds1.2.3dksfj")
+			Ω(ver.GetVersion()).To(Equal("1.2.3"))
+
+			ver.SetFullVersionString("5.2.3  dksfj")
+			Ω(ver.GetVersion()).To(Equal("5.2.3"))
+
+			ver.SetFullVersionString("____ 3.4.3 -dksfj")
+			Ω(ver.GetVersion()).To(Equal("3.4.3"))
```
Amit-PivotalLabs: Realistically, what cases are these assertions trying to test?  When does the CLI version have all these extra characters?

Amit-PivotalLabs:
@simonleung8: Why can't this all be done with three functions:

- shell out to get raw CLI version as a string
- clean raw CLI version to a comparable semantic version
- compare two semantic versions to see if the first is at least the second

Amit-PivotalLabs:
@simonleung8:

I spoke with @jberkhahn, he agrees this can be done much more simply, along the lines of my previous comment.  Note the last step of comparing two semantic versions can be done with other packages, e.g. https://github.com/blang/semver or https://github.com/hashicorp/go-version.

simonleung8:
No problem, will update and make another PR

PR #63, opened 2015-08-19
A slew of changes to make CATs more compatible with Diego

vito:
Also tagged some tests that won't be compatible as of this commit (or any time soon in `cf files`'s case) so we can filter them out via the BOSH errand.

Each commit should provide a bit more detail.

### DIFF
--- a/apps/admin_buildpack_lifecycle_test.go
+++ b/apps/admin_buildpack_lifecycle_test.go
@@ -121,7 +121,7 @@ EOF
 			Expect(err).ToNot(HaveOccurred())
 		})
 
-		It("fails to stage", func() {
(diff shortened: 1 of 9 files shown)


DanLavine:
Hey @vito,

The change to dora's `myip` endpoint breaks a `security_groups/running_security_groups_test` on bosh-lite. Are these changes necessary for Diego or is it just trying to make it easier to read?

Thanks,
@danlavine && @dkoper, CF OSS Release Integration Team

vito:
Those are necessary, not sure why cf would be failing there though. The
idea is to not depend on knowing the interface name and trust what the
platform told the app.

PR #65, opened 2015-09-15
reduce app memory usage to 128M unless previously specified

dliebreich:
Signed-off-by: Jesse Alford <jalford@pivotal.io>

we did the brute force method of decorating all `cf push` with `-m 128M` unless a -m flag was already present – a condition we checked for with manual review.

CATs passes on our environment with the following skip flags set
```
-skipPackage=helpers,v3,security_groups,routing '-skip=SSO|lucid64'
```

please consider merging this in for the cf-218 release.

### DIFF
--- a/apps/admin_buildpack_lifecycle_test.go
+++ b/apps/admin_buildpack_lifecycle_test.go
@@ -109,7 +109,7 @@ EOF
 
 	Context("when the buildpack is detected", func() {
 		It("is used for the app", func() {
-			push := Cf("push", appName, "-p", appPath).Wait(CF_PUSH_TIMEOUT)
+			push := Cf("push", appName, "-m", "128M", "-p", appPath).Wait(CF_PUSH_TIMEOUT)
 			Expect(push).To(Exit(0))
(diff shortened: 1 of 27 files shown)


ryanpei:
@Amit-PivotalLabs we'd really like to see this PR merged so that our CATs will be significantly less flaky. I couldn't find any corresponding story to this PR in the mega backlog, is that the right place to look?

Amit-PivotalLabs:
The story to merge the PR has already been started by the MEGA team.
 cf-gitbot automatically creates stories in the MEGA icebox when a PR is
made against the repo, and comments on the PR with a link to the story,
which you can see above.

ryanpei:
thanks @Amit-PivotalLabs , i completely missed the gitbot comment directly above my comment. i think i'm slowly developing gitbot blindness.

ryanpei:
@Amit-PivotalLabs will this be in the next cf-release by chance?

Amit-PivotalLabs:
Yes

Amit-PivotalLabs:
@ryanpei sorry, I take that back.  I believe we are trying to cut a release today, and it won't be in that SHA.

PR #66, opened 2015-09-15
remove id token usage

fhanik:
API change in the UAA now supports the id_token parameter. CATS doesn't. 

CATS doesn't need the parameter nor use it.


Amit-PivotalLabs:
Looks like this change was committed directly, rather than via PR.  Closing.

PR #68, opened 2015-10-16
logs failures of apps that failed staging [#103840940]

jtarchie:
We've had no insight into some flakey tests and would like to expose more information.

Signed-off-by: JT Archie <jtarchie@gmail.com>


jtarchie:
@Amit-PivotalLabs this would help with evaluate why we have flakiness with a particular PHP fixture. It occurred again -- 1/10 test runs it fails.

PR #71, opened 2015-11-12
Create IPsec test suite

nickwei84:
- Added a test to the IPsec test suite
- Updated the README to include a description of the IPsec test suite
- We add additional api to dora that could return body text with specified size(less than 5 MB), and use it in our new test suite for deployment with IPsec enabled

### DIFF
--- a/apps/large_payload_test.go
+++ b/apps/large_payload_test.go
@@ -1,0 +1,36 @@
+package apps
+
+import (
+	"fmt"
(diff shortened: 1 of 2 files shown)


sykesm:
Hi. Where is there an ipsec in the open source cloud foundry deployment? Should this be part of the open source acceptance test suite?

Thanks.

PR #72, opened 2015-11-12
Add a large payload test to the Applications CATs suite

nickwei84:
- Added an endpoint to the dora app, `/largetext`, that returns a response body of a specified size
- Added a test to the Applications CATs suite to test the return of a large response body from an app.

### DIFF
--- a/apps/large_payload_test.go
+++ b/apps/large_payload_test.go
@@ -1,0 +1,36 @@
+package apps
+
+import (
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


Amit-PivotalLabs:
Hey @nickwei84 

Thanks for the PR.  Can you use the `DEFAULT_MEMORY_LIMIT` constant instead of "128M", or is there an explicit reason for using "128M"?

mmb:
We have changed the 128M to the constant.

- @mmb / @mandyozhao

Amit-PivotalLabs:
This will not clean up after itself property.  `appName` will be empty in the `AfterEach`, because you're shadowing it when setting `appName := ...` in the `It`.

PR #73, opened 2015-12-08
Update app_bits_copy_test.go

sklevenz:
This test fails seldom in case of a slow push (> 30s). 

I assume the DEFAULT_TIMEOUT (30s) shouldn't be used for push. I have changed it to CF_PUSH_TIMEOUT (2 minutes).

### DIFF
--- a/apps/app_bits_copy_test.go
+++ b/apps/app_bits_copy_test.go
@@ -20,8 +20,8 @@ var _ = Describe("Copy app bits", func() {
 		golangAppName = generator.PrefixedRandomName("CATS-APP-")
 		helloWorldAppName = generator.PrefixedRandomName("CATS-APP-")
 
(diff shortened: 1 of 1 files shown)


Amit-PivotalLabs:
Hi @sklevenz 

Since you're saying the test seldom fails for >30s push, is the desired to change it to the 2m timeout really just to use a more sensible constant, i.e. the `PUSH` timeout rather than the plain `DEFAULT` timeout?

sklevenz:
The default for CF_PUSH_TIMEOUT is 2 minutes and this will help in our case. With seldom I mean we have a CF deployment which was fast enough in the past and did not cause issues. But now something changed in infrastructure and push became a little bit slower so that this test fails.

Amit-PivotalLabs:
Okay, sounds reasonable.  Prioritizing.

robdimsdale:
Merged, thanks!

PR #74, opened 2015-12-11
long list of config options is not helpful

shalako:
I found today that the only way to get tests to run is to see them fail, discover a missing config property, run them again, see them fail, add another property, etc. This is a poor experience. If I want to run the `routing` suite, I'd like to go to one place and see concise instructions that will result in a successful run the first time.

Rather than an unsorted list of parameters that aren't always applicable, I'm interested in starting to move documentation of test config options to sections for the specific test suites.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,15 +66,13 @@ EOF
 export CONFIG=$PWD/integration_config.json
 ```
 
-The full set of config parameters is explained below:
+Most config parameters supported by `integration_config.json` are listed here. See setup instructions for specific test suites below for suite-specific parameters.
 
 * `api` (required): Cloud Controller API endpoint.
 * `admin_user` (required): Name of a user in your CF instance with admin credentials.  This admin user must have the `doppler.firehose` scope if running the `logging` firehose tests.
 * `admin_password` (required): Password of the admin user above.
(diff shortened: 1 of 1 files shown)


Amit-PivotalLabs:
Hi, sorry for the delay, still having a think about this one.  There are other configs that are suite-specific, so for completeness and consistency I think we should discover all such config and update things accordingly.  But I'm actually wondering whether the actual right thing to do is to have separate config objects for the suites that require separate config.  It's difficult to maintain up-to-date documentation on all these configs because it's not easy to tell what uses what, which stems from the fact that one massive object is used to configure multiple disparate suites.

ematpl:
+1 for a more modular configuration format in the CATs (and other, similar AT suites). I think part of the problem stems from the config definition living in the [cf-test-helpers](https://github.com/cloudfoundry-incubator/cf-test-helpers/blob/master/helpers/config.go) repo, instead of in the CATs themselves. The lack of cohesion has made it difficult to coordinate updates to or refactors of the configuration schema, and so it's been easiest simply to dump new properties in there, leading to the current sprawl.

Amit-PivotalLabs:
Going to close this PR for now, we'll have to revisit the bigger issues with the massive config and documentation in this repo at a later time.

PR #75, opened 2015-12-14
Update dora README.md with information about /largetext endpoint

dsabeti:
We added an endpoint to Dora for testing large payloads. This change simply adds the endpoint to the list of endpoints in the Dora readme.


robdimsdale:

Rob & @zaksoup

PR #76, opened 2015-12-15
Clean up quota in recursive-delete services test

ematpl:
The original `delete-quota` command here always fails because it is expecting interactive input to confirm the deletion. This commit supplies it with the `-f` flag to avoid the confirmation as well as the admin context authorizing it to perform the delete.


robdimsdale:

Rob & @zaksoup

ematpl:
Thanks, @robdimsdale && @zaksoup!

PR #77, opened 2015-12-17
remove `ruby` directives from Gemfile

jtarchie:
When Ruby versions get deprecated CATs start failing for the wrong reasons.

[#110323846]

Signed-off-by: Brandon Shroyer <bshroyer@pivotal.io>


robdimsdale:

Rob & @zaksoup

PR #78, opened 2015-12-18
remove python version requirement [#110073122]

jtarchie:


robdimsdale:

Rob & @zaksoup

PR #80, opened 2016-01-04
Override default and cf push timeouts if they are configured

mdelillo:
Use configured default and CF push timeouts in place of hardcoded values.


robdimsdale:
Merged, thanks. We will deliver the tracker story once the [corresponding bump in cf-release](https://github.com/cloudfoundry/cf-release/commit/c95ead506cc9c96bb8f34313eb8cb7a27a299fb2) reaches `release-candidate`

Rob && @zaksoup

PR #81, opened 2016-01-16
dats into cats!

cwlbraa:
We moved the relevant dats into the cats. notes can be found here: https://www.pivotaltracker.com/story/show/111125958


onsi:
hurray!

robdimsdale:
merged, thanks!

Rob & @zaksoup & @dsabeti

Issue #87, opened 2016-02-06
some routing tests don't have code to print logs in AfterEach


Amit-PivotalLabs:
Similar to but different from the previous issue, we had another run of our acceptance tests fail, and again had some difficulties debugging due to missing `cf logs` output.  This time, it was due to the `AppReport` function not being called in the `AfterEach` of a couple tests:

* https://github.com/cloudfoundry/cf-acceptance-tests/blob/master/routing/session_affinity_test.go#L144-L150
* https://github.com/cloudfoundry/cf-acceptance-tests/blob/master/routing/session_affinity_test.go#L85-L87

/cc

* cf-acceptance-tests maintainers: @cloudfoundry/cf-release-integration 
* some committers to the routing suite: @crhino, @yuzhangcmu

Issue #86, opened 2016-02-06
routing suite doesn't do cleanup properly


Amit-PivotalLabs:
Our integration tests are currently failing, and it is difficult to debug due to this code:

https://github.com/cloudfoundry/cf-acceptance-tests/blob/d0de1c92551fd8afb636541f1aa8542d97ca7299/routing/routing_suite_test.go#L71-L75

Essentially what happens is that on teardown, we try to fetch the logs for the app in the test given the app's name, but the way the function above works, the app's name is never returned in the case that it fails to push in the first place.  In particular, we're not able to see a time-stamped log of the apps staging and starting processes, to understand why it timed out after 120s.  Here is a snippet of the build output:

```
[2016-02-06 04:45:20.00 (UTC)]> cf push RATS-APP-831035f3-ed92-40f2-562a-40e3a5907d24 -b ruby_buildpack --no-start -m 256M -p ../assets/hello-routing -d a1-app.cf-app.com 
Creating app RATS-APP-831035f3-ed92-40f2-562a-40e3a5907d24 in org CATS-ORG-5-2016_02_06-04h44m58.478s / space CATS-SPACE-5-2016_02_06-04h44m58.478s as CATS-USER-5-2016_02_06-04h44m58.478s...
OK

Creating route rats-app-831035f3-ed92-40f2-562a-40e3a5907d24.a1-app.cf-app.com...
OK

Binding rats-app-831035f3-ed92-40f2-562a-40e3a5907d24.a1-app.cf-app.com to RATS-APP-831035f3-ed92-40f2-562a-40e3a5907d24...
OK

Uploading RATS-APP-831035f3-ed92-40f2-562a-40e3a5907d24...
Uploading app files from: ../assets/hello-routing
Uploading 83.6K, 7 files
Done uploading               
OK


[2016-02-06 04:45:26.02 (UTC)]> cf start RATS-APP-831035f3-ed92-40f2-562a-40e3a5907d24 
Starting app RATS-APP-831035f3-ed92-40f2-562a-40e3a5907d24 in org CATS-ORG-5-2016_02_06-04h44m58.478s / space CATS-SPACE-5-2016_02_06-04h44m58.478s as CATS-USER-5-2016_02_06-04h44m58.478s...
-----> Downloaded app package (512K)
-------> Buildpack version 1.6.12
[...]

PR #88, opened 2016-02-10
Routing CATs fixes for GH issues: #86 and #87

markstgodard:
This PR addresses GH issues both https://github.com/cloudfoundry/cf-acceptance-tests/issues/86 and https://github.com/cloudfoundry/cf-acceptance-tests/issues/87:

- Fix tests so they delete app properly

- Add missing app logging for session affinity tests

Note, talked to Amit and he was ok with a single PR for two issues, since the changes were in same area.

@markstgodard && @chavdarch

### DIFF
--- a/routing/routing_suite_test.go
+++ b/routing/routing_suite_test.go
@@ -68,16 +68,17 @@ func StartApp(app string) {
 	Expect(cf.Cf("start", app).Wait(CF_PUSH_TIMEOUT)).To(Exit(0))
 }
 
-func PushApp(asset, buildpackName string) string {
-	app := PushAppNoStart(asset, buildpackName)
-	StartApp(app)
-	return app
+func PushApp(appName, asset, buildpackName string) {
+	PushAppNoStart(appName, asset, buildpackName)
+	StartApp(appName)
 }
 
-func PushAppNoStart(asset, buildpackName string) string {
-	app := generator.PrefixedRandomName("RATS-APP-")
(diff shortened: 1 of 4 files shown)


routing/routing_suite_test.go
```diff
@@ -68,16 +68,17 @@ func StartApp(app string) {
 	Expect(cf.Cf("start", app).Wait(CF_PUSH_TIMEOUT)).To(Exit(0))
 }
 
-func PushApp(asset, buildpackName string) string {
-	app := PushAppNoStart(asset, buildpackName)
-	StartApp(app)
-	return app
+func PushApp(appName, asset, buildpackName string) {
+	PushAppNoStart(appName, asset, buildpackName)
```
robdimsdale: We should be setting the backend here, between `push --no-start`, and `start`: see https://github.com/cloudfoundry/cf-acceptance-tests/blob/master/README.md#code-conventions

robdimsdale:
@markstgodard && @chavdarch thanks for submitting this PR. We have a comment inline about explicitly setting the backend, and a more general observation that a lot of these helper methods are exported, but they're only used in the same package so maybe they don't need to be?

Let us know if you have any questions, thanks.

ishustava:
Hi @robdimsdale,

Thank you for the feedback. We have fixed that. Please let us know if that works.

@atulkc and @ishustava

APShirley:
Looks good, merging.

- Al & @nimakaviani

PR #89, opened 2016-02-26
Add test to verify container metrics

ScarletTanager:
This commit checks that the container metrics have been submitted for the specified application.

[#113279137]

Signed-off-by: Sandy Cash <scarlet.tanager@gmail.com>

### DIFF
--- a/apps/loggregator_test.go
+++ b/apps/loggregator_test.go
@@ -113,6 +113,33 @@ var _ = Describe("loggregator", func() {
 			Expect(messages).To(ContainElement(EnvelopeContainingMessageLike("Muahaha")), "To enable the logging & metrics firehose feature, please ask your CF administrator to add the 'doppler.firehose' scope to your CF admin user.")
 		})
 	})
+
+	It("shows container metrics", func() {
+		config := helpers.LoadConfig()
+		appGuid := strings.TrimSpace(string(cf.Cf("app", appName, "--guid").Wait(DEFAULT_TIMEOUT).Out.Contents()))
+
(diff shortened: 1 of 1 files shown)


apps/loggregator_test.go
```diff
@@ -113,6 +113,33 @@ var _ = Describe("loggregator", func() {
 			Expect(messages).To(ContainElement(EnvelopeContainingMessageLike("Muahaha")), "To enable the logging & metrics firehose feature, please ask your CF administrator to add the 'doppler.firehose' scope to your CF admin user.")
 		})
 	})
+
+	It(diegoUnsupportedTag+"shows container metrics", func() {
```
ematpl: I don't see why this test doesn't also apply to the Diego backend, as the cells also emit metrics about their running containers periodically.

fraenkel:
It was more a statement that we didn't test it on Diego. If someone can verify there should be no reason why it wouldn't apply to both.

Amit-PivotalLabs:
I agree with @ematpl, let's verify that this works on Diego and remove the tag.

Amit-PivotalLabs:
@ScarletTanager @fraenkel are you folks able to verify this against Diego?  Should be as simple as focusing the test and running the test suite via ginkgo with `$CONFIG` set to a JSON file that specifies `"backend": "diego"` and targets an API with a Diego backend, e.g. BOSH-Lite or A1.

fraenkel:
@Amit-PivotalLabs @ematpl Verified it works on Diego and DEA on A1. Commit has been updated to support both.

robdimsdale:
Merged, thanks.

Rob && @dsabeti

PR #91, opened 2016-03-22
Add protocol to uploadUrl of v3 tests

mtekel:
Add protocol to uploadUrl in all of the v3 tests.

Tests fail without this if your deployment is operating on https instead of http. This is because without protocol specified the tests will try to connect using default protocol and port, which http on port 80.


Amit-PivotalLabs:
@utako @SocalNick can you take a look at this v3-related updated to CATS?

simonleung8:
@mtekel 

Thanks for the PR, could you rebase this PR and fix the conflict?

We will take a look at it once the conflict is fixed.

Thanks
Simon & @petergtz

mtekel:
Done, hopefully you can check it before it needs another rebase...

simonleung8:
Looks good, merged!

Thank you.

PR #93, opened 2016-03-29
Add visual feedback for success and failure of CATs

wendorf:
Signed-off-by: Dan Wendorf <dwendorf@pivotal.io>


bin/test
```diff
@@ -20,3 +20,15 @@ go list github.com/cloudfoundry/cf-acceptance-tests/... | xargs -I {} go test -c
 # Instead, we'll simply not add quotes around $@
 # shellcheck disable=SC2068
 ginkgo $@
+
+err=$?
+if which dishy > /dev/null; then
```
APShirley:
```sh
if which dishy > /dev/null 2>&1; then
```
mo' betta stderr handling (make sure that works first)

PR #96, opened 2016-04-04
Allow firehose test to fail when it times out

### LINKED ISSUE #94: The firehose test doesn't test anything
https://github.com/cloudfoundry/cf-acceptance-tests/blob/master/apps/loggregator_test.go#L87

The for loop at line 104 just has us collecting messages for 5 seconds and bailing.

jasonkeene:

Firehose test was not failing when timeout was hit. It does fail now.

* An explanation of the use cases your change solves:

It properly fails now when timeout occurs, which is useful. This closes #94.

### DIFF
--- a/apps/loggregator_test.go
+++ b/apps/loggregator_test.go
@@ -99,7 +99,7 @@ var _ = Describe("loggregator", func() {
 				return helpers.CurlApp(appName, fmt.Sprintf("/log/sleep/%d", hundredthOfOneSecond))
 			}, DEFAULT_TIMEOUT).Should(ContainSubstring("Muahaha"))
 
-			Eventually(msgChan, 5).Should(Receive(EnvelopeContainingMessageLike("Muahaha")), "To enable the logging & metrics firehose feature, please ask your CF administrator to add the 'doppler.firehose' scope to your CF admin user.")
+			Eventually(msgChan, 5*time.Second).Should(Receive(EnvelopeContainingMessageLike("Muahaha")), "To enable the logging & metrics firehose feature, please ask your CF administrator to add the 'doppler.firehose' scope to your CF admin user.")
 		})
 	})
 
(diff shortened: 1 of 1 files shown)


apps/loggregator_test.go
```diff
@@ -98,19 +99,7 @@ var _ = Describe("loggregator", func() {
...
-				select {
-				case <-timeout:
-					return
-				case msg := <-msgChan:
-					messages = append(messages, msg)
-				}
-			}
-
-			Expect(messages).To(ContainElement(EnvelopeContainingMessageLike("Muahaha")), "To enable the logging & metrics firehose feature, please ask your CF administrator to add the 'doppler.firehose' scope to your CF admin user.")
+			Eventually(msgChan, 5).Should(Receive(EnvelopeContainingMessageLike("Muahaha")), "To enable the logging & metrics firehose feature, please ask your CF administrator to add the 'doppler.firehose' scope to your CF admin user.")
```
robdimsdale: Can we use a `time.Duration` here? As in: `Eventually(msgChan, 5 * time.Second).Should(...`. Not functionally any different, but it's a matter of consistency across the rest of the code base.

dsabeti: Can we use a `time.Duration` here instead of an `int`? As in: `Eventually(msgChan, 5 * time.Second).Shoud(...` It's not functionally any different, but it helps with consistency across the codebase.

dsabeti:
Hey @jasonkeene this mostly looks really good. There was a small note about using `time.Duration` for the timeout. Once that's fixed I'd be happy to merge.

jasonkeene:
@dsabeti Hah, I didn't know that you could give a duration there. Seems a bit verbose. I'll update the PR here in a moment.

robdimsdale:
This looks fine now, thanks. We will merge shortly

PR #103, opened 2016-04-14
Creating buildpack not at 1st position

felixfriedrich:

As buildpacks uploaded by the acceptance_tests do not get properly deleted in case of tests failing, it seems better not to create the buildpack at the first position in the list. But: there are obviously better implementations to achieve what I just described :-)

Apps may use the leftover buildpack in production.

* [ ] I have successfully run these tests against bosh-lite

### DIFF
--- a/apps/admin_buildpack_lifecycle_test.go
+++ b/apps/admin_buildpack_lifecycle_test.go
@@ -103,7 +103,7 @@ EOF
 			_, err = os.Create(path.Join(appPath, "some-file"))
 			Expect(err).ToNot(HaveOccurred())
 
-			createBuildpack := cf.Cf("create-buildpack", BuildpackName, buildpackArchivePath, "0").Wait(DEFAULT_TIMEOUT)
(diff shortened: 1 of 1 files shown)


Amit-PivotalLabs:
Hi @felixfriedrich 

The acceptance tests are not intended for use against production systems, they are intended for use against acceptance environments.  You may wish to use the lighter weight [smoke tests](https://github.com/cloudfoundry/cf-smoke-tests) instead.  Note their description:

> Smoke tests for CloudFoundry that are safe to run in a production environment

felixfriedrich:
Oh, I did not knew that, we usually run both kind of tests against all our environemnts. Thanks for the clarification.

fraenkel:
@Amit-PivotalLabs The intention of this PR is actually correct. If you read the README, it states "should not modify the CF state in such a way as to impact other tests."

Given we are adding a buildpack as #1, we can impact other tests if you aren't specifying the buildpack.

felixfriedrich:
Thanks for having a closer look @fraenkel ! As these tests actually test end user functionality (I see them as integration tests, if that makes sense) I would argue that it even makes sense to run them against production systems. Despite the README does not suggest that this is safe. Doesn't mean though, that it could not be made safe :-)

PR #107, opened 2016-04-29
use go1.5 instead of go1.4

jvshahid:

bump the version of go used by the app.

go buildpack has recently deprecated go1.4. this is causing CATS to fail on CF with the latest buildpack. this change fix CATS

### DIFF
--- a/assets/golang/Godeps/Godeps.json
+++ b/assets/golang/Godeps/Godeps.json
@@ -1,5 +1,5 @@
 {
 	"ImportPath": "go-online",
-	"GoVersion": "go1.4.2",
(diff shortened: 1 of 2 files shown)


robdimsdale:
There was a bit of confusion around this - we merged it manually because we thought the PR didn't have the right version of the SHA of the logging-route-service app. Then we made a second commit to bump the version of the app. The PR would have been fine to merge, but it no longer would have been mergable given our partial merge.

The desired work has all been committed and pushed (and this submodule has been bumped in cf-release) so we will close this issue now.

Apologies for the confusion.

PR #108, opened 2016-05-03
Increase timeoutes for firehose data tests

scottmuc:

We've been experiencing failures that look to be caused by the time to execute the tests are close to the timeouts. We were able to get consistent failures when we dropped the timeout to 15s which leads us to believe that most of the time, the tests are completing at close to the default timeout time span.

We ran this spec against an AWS deployment with the `ginkgo -untilItFails=true` twenty times without failure.

* [x] I have successfully run these tests against ~~bosh-lite~~ an AWS deployed cf installation


DennisDenuto:
Looks good, thanks

PR #111, opened 2016-05-11
Use SleepTimeout instead of DefaultTimeout when sleeping

mdelillo:
This PR changes `time.Sleep` to use `SleepTimeout` instead of `DefaultTimeout`. The CATS suite runs slow for us, so we need to use a high `DefaultTimeout`. Using `SleepTimeout` instead of `DefaultTimeout` lets us have the higher `DefaultTimeout` without sleeping for an unnecessary amount of time.

* [X] I have successfully run these tests against ~~bosh-lite~~ PCF Dev


robdimsdale:
Great, thanks! Merged.

PR #114, opened 2016-05-16
Change file permissions of scp/sftp test asset

jfmyers9:

On linux the `0664` file permission will give write permission to the
current users group. When scping the file into a container, this
permission gets dropped as the group does not exist. On OSX, the group
is not given write permission to the generated file so this is not a
concern. On the errand VM, the vcap group exists in both the container
and the VM, so it is able to transfer the permissions correctly.

[#119334663]

Running the cf-acceptance-tests locally on a linux workstation without the vcap user group.

Signed-off-by: James Myers <jmyers@pivotal.io>


robdimsdale:
Thanks @jfmyers9 - looks good. Merged

PR #116, opened 2016-05-18
Updates loggregator apps test to use latest noaa.

wfernandes:
- Updated dependencies for noaa, dropsonde, sonde-go

[#117935705]

* [ ] I have successfully run these tests against bosh-lite


robdimsdale:
Merged and bumped in cf-release, thanks. We didn't test this on bosh-lite because we don't have the logging infrastructure set up. I assume this is why you didn't do this either.

We will keep an eye on it in the pipeline, and if it causes any issues we will revert the merge.

wfernandes:
I actually ran the `bin/test apps` against our walnut environment. There were some timeout errors but with some other tests. Thought I'd come in today and try to run them against a bosh-lite. If we start seeing legitimate problems with CATs we can revert and investigate.

robdimsdale:
Sounds good - we should have feedback on CATS in A1 in an hour or so. We will let you know if there are any issues

PR #120, opened 2016-06-12
fail fast if no $CONFIG, and print sanitized $CONFIG

Amit-PivotalLabs:
[finishes #121014663]
Related: https://github.com/cloudfoundry/cf-release/issues/972


DennisDenuto:
Looks good. Merging

PR #123, opened 2016-07-06
Update docker test to reflect new env var behavior

jenspinney:
* Previously, $HOME would also be "app" and $TMPDIR would always be "app/tmp"
* With this change, if either of these variables were set with 'cf set-env', we use that value for the variable. Otherwise, we use the value of the variable defined in the Dockerfile, if set. Otherwise, $HOME will be set to the system default (from /etc/passwd) and $TMPDIR will remain unset.

* [X] I have successfully run these tests against bosh-lite 

[#122608871]

Paired with @caod123

### DIFF
--- a/docker/docker_lifecycle_test.go
+++ b/docker/docker_lifecycle_test.go
@@ -87,17 +87,44 @@ var _ = Describe(deaUnsupportedTag+"Docker Application Lifecycle", func() {
 			json.Unmarshal([]byte(env_json), &env_vars)
 
 			By("merging garden and docker environment variables correctly")
+			// values tested here are defined by:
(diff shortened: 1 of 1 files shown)


APShirley:
Thanks @jenspinney looks good, merging.

RochesterinNYC:
@jenspinney @APShirley are there any configuration changes to CF or Diego required to make these tests pass? Our CATs is now failing on these two tests.

jenspinney:
@RochesterinNYC Are you using the latest release of Diego? This should work on v0.1479.0 of Diego and higher.

RochesterinNYC:
We're running `v0.1476.0` so that's probably why. Thanks.

PR #127, opened 2016-07-19
Make sure request only gets one event

zachgersh:

The original events query returned multiple pages but we weren't iterating over all the pages. If your event ended up on the second page the test would fail.

Makes test less likely to fail by looking for the particular event the test specifies and the actee GUID. There should only be one event.

* [X] I have successfully run these tests against a full-scale CF deploy

[#126327439]

Signed-off-by: Ryan Moran <rmoran@pivotal.io>


Amit-PivotalLabs:
I'm gonna move this to the CAPI Tracker, they can assess if this makes sense and merge it in to their tests. /cc @SocalNick @utako

PR #139, opened 2016-10-03
Buildpacks test should respect values set in config.

prysmakou:


dsabeti:
Hi @prysmakou. Thanks for the PR.

Funnily enough, the point of these tests is to test the detection functionality of the platform. Specifically, they assert that, if a user doesn't specify a buildpack name, the platform can determine the correct buildpack to use. So, adding an explicit buildpack parameter to `cf push` would undermine the value of the tests.

Your impulse to make the buildpacks explicit was a good one though: if you look at any other instance of `cf push`, you should see that the buildpack is explicitly used.

Lastly, if your deployment is using customized buildpack names (hence the need to use the buildpack name parameters in the first place), I'd suggest skipping the `detect` suite by setting `"include_detect": false` in your `config.json`.

prysmakou:
Got it. Thanks for the explanation. In my case customer uses pretty outdated and customized ruby and java buildpacks.

PR #144, opened 2016-10-20
Modifies app to return container ip with ip route

genevievelesperance:
Currently dora's method for discovering the container IP is particular to garden. The container networking team would like to change this to be more generic.

https://www.pivotaltracker.com/story/show/132339657

cc @jaydunk


staylor14:
Merged this PR on the CLI.

Issue #142, opened 2016-10-13
droplet bits test cause unexpected error when run in a windows environment


nickwei84:
The following line fails: https://github.com/cloudfoundry/cf-acceptance-tests/blob/master/apps/droplet_uploading_and_downloading.go#L62

The test "Users can manage droplet bits for an app" in the `apps/droplet_uploading_and_downloading.go` runs `tar -z` which in a windows environment will fail to uncompress with `gzip`, and cause the ginkgo tests to exist with status code 2.

Instead of invoking `tar -ztf` to extract the file and list the contents, this could be done in two separate steps (ie. first `gzip` then `tar`). This way it would work on both windows and non-windows environments.

Amit-PivotalLabs:
Thanks @nickwei84 

We don't have Windows machines.  Would you be interested in submitting a PR for this, you would be able to try it on your Windows machine and confirm that it works.

nickwei84:
@amitkgupta Sounds good. Will do.

Amit-PivotalLabs:
Thanks.  Closing issue in anticipation of PR.

PR #145, opened 2016-10-25
updated a test in 'Uploading and downloading droplets' to be compatib…

### LINKED ISSUE #138: CATs v3 helpers introduce a Bash dependency
CATs run on windows in some cases, where there is no guarantee of having working bash. The [v3 helpers use `grep` to pull in the bearer token](https://github.com/cloudfoundry/cf-acceptance-tests/blob/master/helpers/v3_helpers/v3.go#L101) from UAA. 

According to "sources" (I didn't actually speak to them) on the uaa team, currently only `Bearer` type tokens are issued by this endpoint anyway so we may just be able to stop using `bash -c` and `grep`. If this is not the case, it still may be better to handle the output directly in go code so we can keep CATs OS agnostic.

@utako && @SocalNick what do you think about removing the bash code?

nickwei84:
…le with Windows GNUWin32 executables #142
- removed the dependency on 'bash', fixes #138
- use gexec.Session for easier assertions against output stream
- use CF default timeout for gzip and tar

Signed-off-by: Matthew Boedicker mboedicker@pivotal.io

### DIFF
--- a/apps/droplet_uploading_and_downloading.go
+++ b/apps/droplet_uploading_and_downloading.go
@@ -1,7 +1,6 @@
 package apps
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -27,7 +26,6 @@ import (
 
 var _ = AppsDescribe("Uploading and Downloading droplets", func() {
 	var helloWorldAppName string
-	var out bytes.Buffer
 
 	BeforeEach(func() {
 		helloWorldAppName = random_name.CATSRandomName("APP")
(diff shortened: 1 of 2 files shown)


dsabeti:
Thanks @nickwei84 for this PR. Seems reasonable. I'm going to prioritize a story to merge this in.

ChunyiLyu:
Hi @nickwei84.  We can't get this test to pass reliably with your change.  It either passes, or fails waiting for ./tmp or ./logs (line 75 and 76).  Can you take a look to see what you think?

Here's what we see on a typical failed run:
[fail.out.gz](https://github.com/cloudfoundry/cf-acceptance-tests/files/577023/fail.out.gz)

Here is a failed run from our ci:
https://runtime.ci.cf-app.com/teams/main/pipelines/cf-deployment/jobs/acceptance-tests/builds/128

--Steve(staylor14) && Chunyi

nickwei84:
@ChunyiLyu Thanks for the feedback. We'll look into this.
- Cli Team

staylor14:
Hi @nickwei84!  Were you able to figure out what's going wrong?

nickwei84:
@staylor14 Sorry the CLI team hasn't gotten around to doing this yet in our backlog. I'll take a crack at it during my flex hour.

nickwei84:
@ChunyiLyu @staylor14
We submitted a PR #157 to fix the above flake.

keymon:
This PR makes it work on windows, but makes it fail if you have a version of gzip older than 1.6, because the `gzip -k` option does not exist in <= 1.5

We run these tests on a docker container with golang based on debian wheezy, and they bundle gzip 1.5 :(

dsabeti:
Hi @keymon. Is there a specific reason you're using that docker image? We actually maintain a [docker image](https://hub.docker.com/r/relintdockerhubpushbot/cats-concourse-task/) that we suggest for running CATs. If that doesn't suit your needs, I'm happy to discuss making changes to the image or even potentially the CATs repo.

David Sabeti
PM, CF Release Integration

keymon:
Hi, no reason to keep that image, we did not know you had an image. That is cool, I would rather prefer use a supported image. I will try on Monday and come back to you.

PR #147, opened 2016-10-27
Update security groups test for compatibility with container networking

mcwumbly:
This updates the existing security-group tests in CATS for running applications to work with container networking.

It requires a new config option,`include_container_networking` which defaults to `false`

When false, the existing test is run using security groups.  When true, it instead uses the cf cli network policy plugin to create/remove policies.

To work with container networking, this will also require PR https://github.com/cloudfoundry/cf-acceptance-tests/pull/144 be merged.
- Initial work done here for context: [#132951677](https://www.pivotaltracker.com/story/show/132951677)
- Tracking story for this PR: [#133349151](https://www.pivotaltracker.com/story/show/133349151)

/cc @markstgodard

### DIFF
--- a/helpers/config/config.go
+++ b/helpers/config/config.go
@@ -7,6 +7,7 @@ import (
 type CatsConfig interface {
 	GetIncludeApps() bool
 	GetIncludeBackendCompatiblity() bool
+	GetIncludeContainerNetworking() bool
 	GetIncludeDetect() bool
 	GetIncludeDocker() bool
 	GetIncludeInternetDependent() bool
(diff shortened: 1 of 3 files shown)


mcwumbly:
Running the test with `include_container_networking` will require the latest release of the network policy plugin for the cf cli, which you can find here:

https://github.com/cloudfoundry-incubator/netman-release/releases

mcwumbly:
We're running this in our own CI with a [forked and patched version](https://github.com/cloudfoundry-incubator/container-networking-ci/commit/8127021ddda15929dedfad12bf0529ed3548b164) of the [Dockerfile from runtime-ci](https://github.com/cf-container-networking/runtime-ci/blob/master/scripts/ci/run-cats/Dockerfile), so you can see what changes we had to make there:

```Dockerfile
RUN wget -q -O /tmp/network-policy-plugin "https://github.com/cloudfoundry-incubator/netman-release/releases/download/v0.5.0/network-policy-plugin-linux64" && \
  chmod +x /tmp/network-policy-plugin && \
  cf install-plugin /tmp/network-policy-plugin -f && \
  rm -rf /tmp/*
```

PR #148, opened 2016-10-27
Always strip the Expect header from `curl` requests

wendorf:
Some infrastructures (notable Google Cloud Platform when using an HTTP
load balancer) do not allow 100 Continue responses. `curl` will add
'Expect: 100-Continue` to requests that include bodies, prompting
servers to respond with 100 Continue. By unsetting the header, we
prevent 100 Continue responses, allowing the responses to move through
the load balancer.


dsabeti:
Hey @wendorf. This seems like a reasonable change in concept, but this isn't quite the right place to do it. You may have noticed that the file you edited is in the `vendor/` directory, which means that it actually belongs to a dependency repo -- in this case, [cf-test-helpers](https://github.com/cloudfoundry-incubator/cf-test-helpers). If you could submit the PR to that repo, and make sure that all unit tests pass, I'd gladly merge it in.

PR #152, opened 2016-11-16
Use a vendored dotnet core app for detect spec

RochesterinNYC:
- This vendored app is locked to dotnet version 1.0.1, which is LTS
  until ~2019. Any changes in the buildpack should not break this app/spec
  until that LTS ends.
- This reduces the staging/push time for the app drastically and reduces
  the required memory and disk quota

[#134449407]

Signed-off-by: Sam Smith <sesmith177@gmail.com>


RochesterinNYC:
Anything else we need to add to this PR to get it merged ?

PR #154, opened 2016-11-16
Add CATs for Isolation Segments

ScarletTanager:
This PR also adds a few tests to the helpers/config for Isolation Segments. It also removes a duplicate failing test from that folder.


jaresty:
@ScarletTanager We have a couple of issues with this PR that we would like to resolve before pulling it in:

1. The test that was deleted was not redundant; in fact it tests an important error aggregation case so we need to make it work.  We initially saw the same error as you did running under go version 1.6.2 but then updating go to 1.7.4_1 seems to have resolved the failure.  Please try updating go and adding back this test as it is important to us.
2. We noticed that you modified the vendored cf-test-helpers project directly inside of cf-acceptance-tests.  The proper way to make those kinds  of changes is to instead make a pull request against [that separate repository](https://github.com/cloudfoundry-incubator/cf-test-helpers) and then pull in the updated dependency once that PR has been accepted.  Please follow that process if you need to make any changes to cf-test-helpers.

Please update your PR and we will look at it again.  Thank you!

DanLavine:
@jaresty

I pushed a commit to the cf-test-helpers repo and updated the dependency using gvt. I also added back in the removed test and rebased on top of master to get rid of the merge conflicts.

PR #155, opened 2016-11-18
Include protocol in SSO URLs.

jmcarp:
Our api server is behind a proxy that redirects http to https, so making a request to the url without a protocol returns a 301 with an empty body, which panics the sso tests. This patch prepends the protocol to the api url--wdyt?


JesseTAlford:
Seems fine. Have you run CATs and seen it pass with this change?

PR #156, opened 2016-11-18
Pin dotnet version.

jmcarp:
Based on discussion in https://github.com/cloudfoundry/dotnet-core-buildpack/issues/116, it sounds like dotnet apps need to pin a version to deploy correctly.

### DIFF
--- a/assets/dotnet-core/project.json
+++ b/assets/dotnet-core/project.json
@@ -7,7 +7,7 @@
     "Microsoft.AspNetCore.Server.Kestrel": "*",
     "Microsoft.NETCore.App": {
       "type": "platform",
-      "version": "*"
+      "version": "1.0.0"
     },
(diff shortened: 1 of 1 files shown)


degrim:
@jmcarp I would pin version 1.0.1 since that is the latest patch for the LTS release.  Also, a global.json file is required to tell the buildpack to install a specific version of the .NET CLI (and framework).

degrim:
It looks like @RochesterinNYC already submitted a pull request which will resolve the issue in #152 by using a vendored app, that solution seems like a more permanent fix.

dsabeti:
Hi @jmcarp. @degrim is correct, I think the other PR covers this. I'm going to close this PR, but feel free to re-open it if you think that 152 doesn't address your concern.

PR #159, opened 2016-11-28
update "can delete the key" service test in service_instance_lifecycle.go

nickwei84:
This test recently broke because of a change in behavior to the "service-key" command - as a result of a bug fix in the CLI (tracker story #134271615).

In short, when running the "service-key" command, it used to exit 0 when the key is not found. It now exits 1. This PR updates a test assumption accordingly.

### DIFF
--- a/services/service_instance_lifecycle.go
+++ b/services/service_instance_lifecycle.go
@@ -212,10 +212,7 @@ var _ = ServicesDescribe("Service Instance Lifecycle", func() {
 							Expect(deleteServiceKey).To(Exit(0), "failed deleting service key")
 
 							keyInfo := cf.Cf("service-key", instanceName, keyName).Wait(Config.DefaultTimeoutDuration())
-							Expect(keyInfo).To(Exit(), "failed key info")
-
-							errorMessage := fmt.Sprintf("No service key %s found for service instance %s", keyName, instanceName)
-							Expect(keyInfo).To(Say(errorMessage))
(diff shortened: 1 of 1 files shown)


dsabeti:
@nickwei84 It looks like the new assertion doesn't assert on an exit code, which makes it a less meaningful test. Would it make sense to say `Expect(keyInfo).To(Exit(1), "failed key info")` instead?

nickwei84:
@dsabeti By asserting an exit code of 1 for the `service-key` command, we will make this test incompatible with older versions of the CLI.

You're right this change makes the test less meaningful. We're rationalizing the absence of this assertion on the `service-key` command because the test is concerned with the `delete-service-key` command, and we are testing the exit code there.

dsabeti:
@nickwei84 Ok makes sense. In that case, you can actually get rid of the expectation altogether, since `.Wait()` already asserts that the process exits. And if we don't make an assertion on exit code, can we at least assert that the output includes the appropriate error message? It would look something like this:
```
Expect(keyInfo).Say("No service key X found for service instance Y"))
```
I just want to make sure that we don't lose coverage here.

nickwei84:
@dsabeti sorry for the slow response. The above commit removes the duplicate assertion on session exit and simplifies the code. It is still asserting on the error message that the deleted service key cannot be found.

JesseTAlford:
Sorry for our slow response back, @nickwei84! We're merging this now.

PR #166, opened 2016-12-14
Remove Evenutally in favor of Expect

zachgersh:
- this test has been extremely flakey due to
  gbytes say not showing up in the requisite amount
  of time.
- no reason to actually use an eventually as all of
  the logs have already been fetched on the line above
  and we are not waiting for them to "Eventually" come
  in.

Signed-off-by: Zachary Gershman <zgershman@pivotal.io>

### DIFF
--- a/routing/zipkin_tracing.go
+++ b/routing/zipkin_tracing.go
@@ -66,7 +66,7 @@ var _ = ZipkinDescribe("Zipkin Tracing", func() {
 
 				appLogsSession = cf.Cf("logs", "--recent", hostname)
 
-				Eventually(appLogsSession.Out).Should(gbytes.Say("x_b3_traceid:\"fee1f7ba6aeec41c"))
+				Expect(appLogsSession.Out).Should(gbytes.Say("x_b3_traceid:\"fee1f7ba6aeec41c"))
 				_, appLogSpanId, _ := grabIDs(string(appLogsSession.Out.Contents()), traceId)
 
 				Expect(curlOutput).To(ContainSubstring(traceId))
--- a/routing/zipkin_tracing.go
+++ b/routing/zipkin_tracing.go
@@ -13,10 +13,10 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gbytes"
+	"github.com/onsi/gomega/gexec"
 )
 
 var _ = ZipkinDescribe("Zipkin Tracing", func() {
-
 	var (
 		app1              string
 		helloRoutingAsset = assets.NewAssets().SpringSleuthZip
@@ -32,7 +32,6 @@ var _ = ZipkinDescribe("Zipkin Tracing", func() {
 
 	AfterEach(func() {
 		helpers.AppReport(app1, Config.DefaultTimeoutDuration())
-
 		helpers.DeleteApp(app1, Config.DefaultTimeoutDuration())
 	})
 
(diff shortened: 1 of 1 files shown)


zachgersh:
@dsabeti you had a chance to look at this?

dsabeti:
Hey @zachgersh, thanks for pinging me. Have you seen better performance with `Expect`? I'd be surprised, because the session (`appLogsSession`) hasn't performed a `.Wait()` yet. Typically, for any `gexec.Session`, you need to either `.Wait()` or assert with `Eventually`.

zachgersh:
@dsabeti isn't this call blocking anyway?

```go
appLogsSession = cf.Cf("logs", "--recent", hostname)
```

You have to wait for that function to return and it doesn't stream logs but just gets the most recent. It has already finished so there is no async component that would need a wait?

Amit-PivotalLabs:
cf.Cf is non-blocking.  It just creates a *gexec.Session.

zachgersh:
@Amit-PivotalLabs thanks - the wait probably should be implicit on the finish of getting all of the logs instead of just waiting for a long time to appear in the next section. Would be nice if calls to `logs --recent` did block, that call should be quick to return.

dsabeti:
Well, `cf.Cf` doesn't know if the command it's running (in this case, `logs --recent`) is blocking or not. All it knows is that it's shelling out to the cf CLI and gets a handle to the session. It's up to the caller to know whether or how long to wait. Also, consider `cf push` -- it's "short-lived" in that it eventually terminates execution, but it will probably take a pretty long time. `cf.Cf` is non-blocking by design, so that you can run any kind of cf command with it.

zachgersh:
if it returns a `gexec.Session` I can change this to wait until it exits `0` and then proceed. That log call SHOULD NOT take that long and it will be way less flakey of a test.

Amit-PivotalLabs:
@zachgersh I agree, but also the `Wait` should respect the `$CONFIG`-provided timeouts, so that people in environments where `cf logs --recent` is slow can do something about it.  Might make sense to recommend this in general in the repo's [Code Conventions](https://github.com/cloudfoundry/cf-acceptance-tests#code-conventions).

dsabeti:
@zachgersh I also agree. Feel free to re-submit the PR and I'll prioritize it.

zachgersh:
Hey @dsabeti / @Amit-PivotalLabs went ahead and switched to:

`Eventually(appLogsSession, Config.DefaultTimeoutDuration()).Should(gexec.Exit(0))`

I also refactored the `grabIDs` method while I was in there. I'd want to run this again but to me it appears to have been simplified in addition to keeping its original behavior.

Please let me know if you have other feedback!

zachgersh:
@dsabeti poke for review :D

dsabeti:
Hey @zachgersh, thanks for the reminder. I just prioritized the story, so the team should get to it soon.

JesseTAlford:
This looks fine and we've merged it. What I don't understand, @zachgersh, is how this makes the test any more reliable in an absolute sense. If it timed out before, it will still time out.

It _does_ simplify a helper function, and that's good. And it clarifies that the problem is that the log fetch didn't exit in the event that it times out, not that the match never arrived (or that the log line was missing, as the case may be). Again, clarity is good.

It also makes the timeout configurable - is this the thing that is going to make it less flakey, that it can respect the timeout configuration? Or I missing something about how this will make the test less flakey in the sense of failing intermittently for performance reasons?

zachgersh:
@JesseTAlford - the change I made was this:

```
Eventually(appLogsSession, Config.DefaultTimeoutDuration()).Should(gexec.Exit(0))
```

from this

```
Eventually(appLogsSession.Out).Should(gbytes.Say("x_b3_traceid:\"fee1f7ba6aeec41c"))
```

Waiting for the `logs --recent` command to finish and then parsing them should make them less flakey (in addition to putting the configurable timeout on session exit). Also note that the original parsing code had the default timeout for `Eventually`. Also, this should present a better error if the session doesn't consistently exit.

JesseTAlford:
Yeah, I read the code, I think I understand the change, all the benefits you mention are real.

My one, specific confusion is the assertion that at waiting for the command to finish and then parsing will make it less flakey. It makes me feel I'm missing something. If it was timing out before on parsing the output, it will time out now on waiting for the command to exit, right? Or it will fail faster if the real problem is that the output is just not there. Either way, failure rate is the same for a given timeout, right?

zachgersh:
It shouldn't timeout waiting for the command to exit based on adding the configurable default timeout. The fix is twofold - move to more specifically waiting for the command to finish and apply the configurable timeout (which was not applied to log parsing). Those two bits in concert make it less flakey (and more explicit as you mention).

JesseTAlford:
Cool, I think we understand each other well enough (and maybe have all along). Why the fix is good is definitely not what I'm working to understand. I'm working to understand how switching up where the Eventually is improves reliability. It sounds like it doesn't, but that's okay because you're not _trying_ to improve reliability with that part of the change, but clarity.

I'm probably misparsing "flakey," a term I find leads to this kind of communication issue pretty readily. I meant it as something like "fails in a way that doesn't signify a defect, generally for reasons of infrastructure or poor time-related test design." Adding the configurable timeout _deffo_ addresses this. Making it clearer exactly what failed by waiting for the command to exit instead of waiting _the exact same amount of time_ on the live output doesn't, to the extent that "flakeyness" is only what I described above.

To the extent that "flakey" actually means "fails for obscure reasons in a way we tend to ignore," then both address it.

Thanks for talkin' this through with me even though we already merged your change and it's probably a minor semantic point.

PR #175, opened 2017-01-17
test app start events through cf-cli

zachgersh:
- no more parsing logs directly from the cc

There is a story in the capi backlog for dis.

Signed-off-by: Zachary Gershman <zgershman@pivotal.io>

### DIFF
--- a/v3/app_lifecycle.go
+++ b/v3/app_lifecycle.go
@@ -14,6 +14,8 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
+	"github.com/onsi/gomega/gexec"
 )
 
 var _ = V3Describe("v3 buildpack app lifecycle", func() {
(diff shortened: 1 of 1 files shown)


dsabeti:
@SocalNick @utako This PR seems good to me, but it's in the v3 suite and think that CAPI should weigh in before we merge. What are your thoughts on merging this in?

SocalNick:
I believe this was discussed with CAPI before submission or shortly after. @utako @aashah can you let @dsabeti know when this is good to merge?

zrob:
@zachgersh Unfortunately this changes the intention of the test.  The previous test was checking for AppUsageEvents across multiple processes, which is ensuring that each process can be properly billed.

The new implementation is testing AuditEvents. These are used as an audit trail.

I think it is important to continue ensuring that we generate the proper billing information.  Unfortunately there is not a cli friendly way that I'm aware of to list the usage information, which makes the whole page issue a problem still :(

dsabeti:
In light of @zrob's comment, it sounds like we should close this PR. 

@zachgersh, any opposition to me closing this? Was this just a good citizen refactor, or is there a specific need not being met? In the latter case, could you submit a different PR?

zachgersh:
@dsabeti I'd actually like to fix this - the test is flakey with the way it looks up events so I'd love to figure out a better way to do that. Maybe @zrob has an idea on how to do that.

dsabeti:
@zachgersh Sure, seems reasonable to make tests less flaky.

Can you describe the failure mode of the flakiness? Maybe we can find another way to make the tests more reliable. Would it help to add something like retry logic to [LastPageUsageEvents](LastPageUsageEvents)?

zrob:
I'd still like to see this resolved, I've opened an issue for the underlying problem, you can drop it in the capi backlog @dsabeti

dsabeti:
@zrob I just transferred the story to the CAPI backlog

zrob:
Closing in favor of https://github.com/cloudfoundry/cf-acceptance-tests/issues/184

PR #178, opened 2017-01-20
Do not use gzip -k option for compatibility

keymon:
Note: Follow up of the comment here: https://github.com/cloudfoundry/cf-acceptance-tests/pull/145#issuecomment-274111140

Older versions of gzip =< 1.5 do not support the -k option.

This is a problem, for instance,  when running the cf-acceptance-tests
using a golang docker image based on debian wheezy, which bundles gzip 1.5.

Here we refactor the test to uncompress and compress again the files,
simulation the behaviour of gzip -k. This change should be compatible
across multiple OS versions.


dsabeti:
Hi @keymon. Can I close this PR since we've resolved the issue: https://github.com/cloudfoundry/cf-acceptance-tests/pull/145#issuecomment-274111140?

keymon:
Sure, I will, it seems to be working fine for us.  

Thank you.

PR #183, opened 2017-02-03
Skip ContainerNetworking needed test in beforeEach

michaelxupiv:
- add skip message to skip_messages

### DIFF
--- a/security_groups/running_security_groups.go
+++ b/security_groups/running_security_groups.go
@@ -21,6 +21,7 @@ import (
 	"github.com/cloudfoundry/cf-acceptance-tests/helpers/app_helpers"
 	"github.com/cloudfoundry/cf-acceptance-tests/helpers/assets"
 	"github.com/cloudfoundry/cf-acceptance-tests/helpers/random_name"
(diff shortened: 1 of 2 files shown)


security_groups/running_security_groups.go
```diff
@@ -169,6 +170,10 @@ var _ = SecurityGroupsDescribe("Security Groups", func() {
 		var clientAppName, securityGroupName string
 
 		BeforeEach(func() {
+			if !Config.GetIncludeContainerNetworking() {
+				Skip(skip_messages.SkipContainerNetworkingMessage)
+			}
+
```
jaresty: Can you move this conditional up the the BeforeEach for the SecurityGroupsDescribe that starts on line 161?  That would prevent pushing any apps.  The test is still pushing an app and takes 90 seconds to fail on my machine.

jaresty:
@michaelxupiv Can you please make this change?  If not, would it be OK with you if we make it for you?

michaelxupiv:
@jaresty feel free! Go ahead sorry for the late response

PR #188, opened 2017-03-15
v3 helpers retrieves doppler endpoint from v2/info

keymon:
The v3 helper has building the doppler URL in a hardcoded way. It does
not include the protocol, so it assumes HTTP, does not add the port and
assumes that the name is s/api/doppler/

In this commit we query the /v2/info endpoint to extract the doppler
endpoint. We replace the websocket protocol (ws) with HTTP (http). If
it is a SSL Websocket (wss) the string replacement will result in https.

In our specific case, we only serve HTTPS for doppler, so the test fails due timeout.

### DIFF
--- a/helpers/v3_helpers/v3.go
+++ b/helpers/v3_helpers/v3.go
@@ -172,7 +172,7 @@ func AssignDropletToApp(appGuid, dropletGuid string) {
 }
 
 func FetchRecentLogs(appGuid, oauthToken string, config config.CatsConfig) *Session {
-	loggregatorEndpoint := strings.Replace(config.GetApiEndpoint(), "api", "doppler", 1)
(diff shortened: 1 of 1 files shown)


dsabeti:
Hey @keymon. Could you explain a little more why you need to replace the Websocket protocol with `http`?

Also, I'll ask @zrob and @Gerg to take a look at this PR. They're have more experience with the `v3` test suite.

keymon:
The original test shells out to call curl, and curl uses http, no websocket. In the original code you do not specify the protocol, so implicitly uses http

dsabeti:
I see, @keymon. You're just replicating the existing behavior.

@zrob Should I prioritize this in the RelInt backlog, or would you rather have CAPI look at it?

ScarletTanager:
@ericpromislow and I ran CATS on this, we're merging this PR.

Issue #184, opened 2017-02-07
V3 App Lifecycle Test can miss app usage events with high concurrency


zrob:
From: #175 

The issue is that with high enough concurrency, just looking at a single page of app usage events might not include the event we are looking for.  The proposed PR has an issue in that it uses the cli to look for audit events, when what we want to test are app usage events.  

Can we look through some limited subset of pages?  Can we increase the number of events per page?

zrob:
@zachgersh @rainmaker this should be fixed as of https://github.com/cloudfoundry/cf-acceptance-tests/pull/193

If you get a chance to crank up your concurrency at let us know how things go that'd be groovy.

Sorry it took so long to get to!

PR #193, opened 2017-03-27
Improve assertions against app usage events

jenspinney:
When querying for recent app usage events, follow pages so that we
won't miss an event due to high activity.

[#139323913](https://www.pivotaltracker.com/story/show/139323913)

Addresses issue #184 and pr #175

Signed-off-by: Aakash Shah <ashah@pivotal.io>


jenspinney:
(I closed and reopened the PR to make cfdreddbot happy)

PR #194, opened 2017-03-28
test syslog drain for matching regex

jtarchie:
The test would always blow up if the first `cf logs --recent` did not match the regex. This ensures that `Eventually` will continue and fail gracefully.

Signed off by: Eno Compton <ecompton@pivotal.io>

### DIFF
--- a/apps/syslog_drain.go
+++ b/apps/syslog_drain.go
@@ -76,14 +76,13 @@ var _ = AppsDescribe("Logging", func() {
 func getSyslogDrainAddress(appName string) string {
 	var address []byte
 
-	Eventually(func() []byte {
+	Eventually(func() [][]byte {
 		re, err := regexp.Compile("ADDRESS: \\|(.*)\\|")
 		Expect(err).NotTo(HaveOccurred())
 
 		logs := cf.Cf("logs", appName, "--recent").Wait(Config.DefaultTimeoutDuration())
-		address = re.FindSubmatch(logs.Out.Contents())[1]
-		return address
-	}, Config.DefaultTimeoutDuration()).Should(Not(BeNil()))
+		return re.FindSubmatch(logs.Out.Contents())
+	}, Config.DefaultTimeoutDuration()).Should(HaveLen(2))
 
 	return string(address)
 }


apps/syslog_drain.go
```diff
@@ -76,14 +76,13 @@ var _ = AppsDescribe("Logging", func() {
...
 		Expect(err).NotTo(HaveOccurred())
 
 		logs := cf.Cf("logs", appName, "--recent").Wait(Config.DefaultTimeoutDuration())
-		address = re.FindSubmatch(logs.Out.Contents())[1]
-		return address
-	}, Config.DefaultTimeoutDuration()).Should(Not(BeNil()))
+		return re.FindSubmatch(logs.Out.Contents())
+	}, Config.DefaultTimeoutDuration()).Should(HaveLen(2))
 
 	return string(address)
```
dsabeti: @jtarchie I'm pretty sure this `address` variable will always be an empty byte array. When do you actually assign a value to it?

jtarchie: Take a look [here](https://play.golang.org/p/eS4AtB3UbE). If the match fails, the return value of `re.FindSubmatch(logs.Out.Contents())` will be an empty slice. If the match succeeds, the return value will be a slice with two members, i.e., the matching string and the capture group, both in the form a slice of bytes.

dsabeti: Sure, but in the happy path, where do you take the matched string and assign it to `address` variable?

The return value of `FindSubmatch` is give to the `return` of the `Eventually` closure, which is used (correctly) against the `HaveLen` matcher. But it never gets assigned to `address`. I'm pretty sure you want something like:
```
var match [][]byte
Eventually(...
  ...
  logs := cf.Cf("logs", appName, "--recent").Wait(Config.DefaultTimeoutDuration())
  match = re.FindSubmatch(logs.Out.Contents())
  return match
}, ...).Should(HaveLen(2))
address = match[1]
```

flawedmatrix:
FWIW, I have a pull request to fix the issue of the tests panicking when `cf logs --recent` does not match the regex:

https://github.com/cloudfoundry/cf-acceptance-tests/pull/192

dsabeti:
Hey @flawedmatrix. Thanks for catching this duplication. I definitely should have recognized that the two PRs were trying to address the same issue. I'll make sure that the engineering team considers both PRs at the same time and does what makes the most sense.

jtarchie:
Closing my PR in favor of @flawedmatrix

PR #202, opened 2017-05-01
Update go fixture apps to use default go version

sesmith177:
- by using go native vendoring and not specifying GOVERSION, the go
buildpack will use the default version of Go. This should future-proof
these fixtures against versions of Go being removed from the buildpack

[#142416247]

Signed-off-by: Anna Thornton <athornton@pivotal.io>
Signed-off-by: Nikita Rathi <nrathi@pivotal.io>

### DIFF
--- a/apps/healthcheck.go
+++ b/apps/healthcheck.go
@@ -10,6 +10,8 @@ import (
 	"github.com/onsi/gomega/gbytes"
 	. "github.com/onsi/gomega/gexec"
 
+	"path/filepath"
+
 	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
 	"github.com/cloudfoundry-incubator/cf-test-helpers/helpers"
 	"github.com/cloudfoundry/cf-acceptance-tests/helpers/random_name"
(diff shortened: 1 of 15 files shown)


dsabeti:
Hey there @sesmith177. This seems like a reasonable fix and makes our lives easier going forward. One comment: I see that we now how to point all `-f` flags to the manifest file: would it make sense to update the [code for managing assets](https://github.com/cloudfoundry/cf-acceptance-tests/blob/master/helpers/assets/assets.go#L39) to point to the manifest file?

sesmith177:
Hello @dsabeti 

If we pointed NewAssets().Golang at `assets/golang/manifest.yml` (and similarly for the other go fixtures), then we would have to remove all the `-p` flags when pushing those apps and specify the path in the `manifest.yml`, which would work

However, I think it might be confusing if sometimes the values in the Assets struct pointed to directories, and sometimes pointed to manifest files

dsabeti:
Makes sense. I'll prioritize the story to merge this.

PR #210, opened 2017-06-01
Fix ordering problem in afterEach

matt-royal:
Before, if the `cf delete-org` in the spec failed, then the afterEach
would fail to delete the quota.

Signed-off-by: Tim Downey <tdowney@pivotal.io>

```
cf delete-quota -f CATS-1-QUOTA-5baada7e-b579-45c2-7 
Deleting quota CATS-1-QUOTA-5baada7e-b579-45c2-7 as admin...
FAILED
Please delete the organization associations for your quota definition.
```


matt-royal:
We accidentally pushed this change to master. Once we realized our mistake, we pushed a revert commit. We'll submit another pull request with the reverted-revert to bring it back in.

PR #213, opened 2017-06-02
Fix default environment variables #146531275

tcdowney:
**WARNING**: Do not merge until `cf-deployment` master is using `capi-release` `1.29.0` or above.

Previously these tests would panic with the following message if log output from staging didn't match exactly what the tests expected:

```
Test Panicked
    runtime error: index out of range
```

We've removed the need for it to look for `START_ENV_CMD` and `END_ENV_CMD` in the output so it should be a little more reliable and will output more meaningful failure messages.

- @ljfranklin && @tcdowney

### DIFF
--- a/apps/default_environment_variables.go
+++ b/apps/default_environment_variables.go
@@ -18,12 +18,10 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


dsabeti:
Ok, it looks like `master` has capi-release 1.30. I'll prioritize the story to merge this PR.

PR #225, opened 2017-07-14
Cleanup leaky isolation segment relationships

anniesing:
We don't have an environment that is setup to run the test at `routing_isolation_segments/routing_isolation_segments.go:103`, so we haven't been able to run this test, but we believe it's good.

[#145311439](https://www.pivotaltracker.com/story/show/145311439)

-- @anniesing && @mikexuu

### DIFF
--- a/helpers/v3_helpers/v3.go
+++ b/helpers/v3_helpers/v3.go
@@ -109,19 +109,6 @@ func CreateIsolationSegment(name string) string {
 	return isolation_segment.Guid
 }
 
-func CreateOrGetIsolationSegment(name string) (string, bool) {
-	var isoSegGuid string
-	var created bool
-	if IsolationSegmentExists(name) {
-		isoSegGuid = GetIsolationSegmentGuid(name)
-		created = false
-	} else {
-		isoSegGuid = CreateIsolationSegment(name)
-		created = true
-	}
-	return isoSegGuid, created
-}
-
 func CreatePackage(appGuid string) string {
 	packageCreateUrl := fmt.Sprintf("/v3/packages")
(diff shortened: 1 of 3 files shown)


dsabeti:
Hi @anniesing @mikexuu, could you say a little more about what's going on in this PR? What resources are leaking, and under what circumstances? How does the PR prevent those resources from getting leaked? Based on the the Tracker story you linked to, it seems like there's an issue with this test when using a persistent org. Does this change allow you to use either kind (persistent or ephemeral) org?

anniesing:
Hi @dsabeti, yes you are correct in that there's an issue when the tests are using a persistent org.

Turns out that we were never properly cleaning up the isolation segment relationship with the org in each after each of the isolation_segments test or routing_isolation_segments. Therefore, the isolation segment relationship is leftover on the persisted org after the sits run. This is because revoking the org entitlement for isolation segments in the after each doesn't work unless you unset the default isolation segment first. Once we're done revoking the entitlement, we then have to delete the isolation segment or else that continues persisting after the test is done running.

This PR addresses those problems and correctly cleans up the created isolation segment and relationships by unsetting the default isolation segment for the persisted org, revoking entitlement, and deleting the isolation segments that are created for the purpose of these tests.

jaresty:
We ran this PR and it did not work at first.  So, we had to make some changes:

- It now will always clean up the fake isolation segments created by some tests
- It will always clean up the all test isolation segments' relationships with their orgs
- It will create the specified isolation segment if it does not exist; otherwise it uses the existing one

The isolation segment specified in the config is NOT deleted since deleting results in a race condition between nodes that are attempting to operate on the isolation segment in parallel (and since they all use the same name that was defined in the deployment manifest, there is no simple solution for this)

PR #227, opened 2017-07-21
Add Service Broker Schema tests

ablease:
As a service broker author, I can register a single plan with a create service instance schema [#145580729](https://www.pivotaltracker.com/story/show/145580729)

As a client tooling author, I can retrieve a create service instance schema for a plan [#146453595](https://www.pivotaltracker.com/story/show/146453595)

## Why

The Open Service Broker API is [proposing](https://github.com/openservicebrokerapi/servicebroker/issues/59) allowing brokers to define JSON schema for their configuration parameters. This will allow tooling to validate parameters and UIs to auto generate forms.

Schemas are to be defined as part of the catalog on a plan and support create/update parameters on a service instance and create parameters on a service binding (update does not exist yet).

The updated spec can be found [here](https://github.com/avade/servicebroker/blob/method-attached-schemas/spec.md#schema-object).

Example of what a new catalog with schemas will look like:
```
{
     ....
    "plans": [{
      "name": "fake-plan-1",
      "id": "d3031751-XXXX-XXXX-XXXX-a42377d3320e",
      "description": "Shared fake Server, 5tb persistent disk, 40 max concurrent connections",
      "schemas": {
        "service_instance": {
          "create": {
            "parameters": {
              "$schema": "http://json-schema.org/draft-04/schema#",
              "type": "object",
              "properties": {
                "billing-account": {
                  "description": "Billing account number used to charge use of shared fake server.",
                  "type": "string"
                }
              }
            }
          }
        }
      }
    }]
}
```

## What

This PR adds CATS for create instance schemas.
1. Test default schema for brokers that do not provide them.
1. Test brokers that provide schemas.

## Notes
For now this feature is only exposed through the CC API. So this PR uses `cf curl`. In the future, this should be replaced with the appropriate CF Cli command.
The feature is not complete in CAPI yet, there will be more PRs into CATS for additional tests down the line.

Feedback appreciated!

## PR 

Sam & @ablease

cc: @mattmcneeney @st3v

### DIFF
--- a/assets/service_broker/cats.json
+++ b/assets/service_broker/cats.json
@@ -51,7 +51,8 @@
                       "content": "40 concurrent connections"
                     }
                   ]
-                }
+                },
+                "schemas": "<fake-plan-schema>"
               },
               {
                 "name": "<fake-plan-2>",
--- a/helpers/services/broker.go
+++ b/helpers/services/broker.go
@@ -22,8 +22,17 @@ import (
 )
 
 type Plan struct {
-	Name string `json:"name"`
-	ID   string `json:"id"`
+	Name    string      `json:"name"`
+	ID      string      `json:"id"`
+	Schemas PlanSchemas `json:"schemas"`
+}
+
+type PlanSchemas struct {
+	ServiceInstance struct {
+		Create struct {
+			Parameters map[string]interface{} `json:"parameters"`
+		} `json:"create"`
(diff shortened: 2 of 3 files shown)


dsabeti:
I'm going to close this PR in favor of https://github.com/cloudfoundry/cf-acceptance-tests/pull/229

PR #239, opened 2017-09-25
Wrap the callback in a describe in order to avoid AfterEach invocations

dprotaso:
When running no tests this dropped the time from 45s to ~20s (bosh-lite)

Signed-off-by: Vikram Yadav <vyadav@pivotal.io>


selzoc:
this is an awesome change!

PR #241, opened 2017-09-28
Add test for CredHub-enabled service keys

tcdowney:
- Updates the Credhub Service Broker to take in a SERVICE_NAME
  environment variable that allows it to have a CATS-specified name for
  its service offering
- Also updates the Credhub Enabled App to look for this same env var to
  properly read the service from its VCAP_SERVICES env
- Credhub Service Broker can now create CredHub credentials with both
  mtls-app and uaa-client actors

[#150753759](https://www.pivotaltracker.com/story/show/150753759)

### DIFF
--- a/assets/credhub-enabled-app/README.md
+++ b/assets/credhub-enabled-app/README.md
@@ -1,0 +1,4 @@
+To rebuild credhub-enabled-app.jar:
+```bash
+./gradlew build jar
+```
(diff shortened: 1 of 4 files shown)


ebeer:
Looks good! :shipit:

tcdowney:
`capi-release` 1.43.0 is required for the new service-key test to pass

PR #244, opened 2017-10-16
Allow task to be running or succeeded

SocalNick:
- Smaller environments might be too slow to catch the running state


selzoc:
👍

PR #265, opened 2018-02-13
Update credhub-service-broker to use UAA client authentication

benmoss:
This is a PR to change the service broker we are using to test CredHub to write using UAA client credentials instead of MTLS.

Write access is going to be removed from MTLS-based authentication in future versions of CredHub, and brokers will be required to get UAA client credentials instead.


ebeer:
Additional CredHub integration test changes:
Our service broker test fixture now uses UAA client authentication for writing to CredHub, which is consistent with planned integration changes in CredHub.
Modified config so that CredHub location is not required in a running-environment-variable-group, but instead can be set via config and defaults appropriately.

ishustava:
Hey @benmoss @ebeer,

We've looked at the PR and would like to suggest a few changes and ask some questions:
- There are a lot of files `vendor`ed in to the `credhub-service-broker` asset. We found that running `dep prune` from that directory removes ~900 of unused files. 
- Integration config now requires a client for the CredHub service broker. This means that there is now a manual step required after a CF deploy to create this client in UAA, which is not ideal. Would it make sense to instead pass a UAA admin user for CATs and for the CredHub tests to then create a service broker client?
- Are you expecting any other values for `credhub_location`, except for the internal URL? Maybe we can omit that property from the integration config entirely? Also, the description for this property mentions PAS, which is a Pivotal proprietary product.

benmoss:
@ishustava since it's possible someone might want to run this without bosh DNS enabled, how about we leave the `credhub_location` configurable?

With the UAA change, it seems like this is going to make the CredHub CATs even harder to run, you'll now need to provide the `uaa_location`, the `uaa_admin`, and the `uaa_admin_secret`. Is that going to be easier than just creating a UAA user for CredHub? There is already a UAA client with `credhub.read` and `credhub.write` in cf-deployment, `cc_service_key_client`. Can you just default to grabbing those credentials?

dsabeti:
Hey @benmoss. Thanks for this feedback. 

I originally missed the fact that `cc_service_key_client` had sufficient scopes for Credhub. In that case, can you make `cc_service_key_client` the default value for `credhub_client`? On our side, we can update our [update-runtime-configs task](https://github.com/cloudfoundry/cf-deployment-concourse-tasks/tree/master/update-integration-configs) to pull the corresponding secret from the vars-store, so that anybody who wants to run these tests can pretty easily.

As for `credhub_location`, is there anybody currently using an external Credhub url? If so, then I'm happy to add it (otherwise, I'd rather wait for someone to ask for it specifically).

ebeer:
@dsabeti We can absolutely put in that default value for `credhub_client`

Concerning `credhub_location`: The previous iteration of this test would fail immediately if CredHub's location was not in the running-environment-variable-group prior to starting the test suite.  Our goal with this change was to make this test easier to run and not require an extra bit of setup in the deployment just for the test. No one is yet using an external CredHub url because it is not possible in the current configuration, but providing that capability is on the roadmap. 

If we pull `credhub_location` from the config, I'd like to respect the  running-environment-variable-group setting as an override to the default value if that sounds reasonable to you.

dsabeti:
Hey @ebeer. Thanks for the context around `credhub_location`. Now that I know it's an alternative to the running environment variable, I'm happy to take the change to include it in the integration config instead.

Let us know when `credhub_client` has a default, and I'll prioritize this for merging. Thanks for your patience while we gained context on this change!

ebeer:
@dsabeti done!

PR #270, opened 2018-02-21
Add wats tests to cats

sunjayBhatia:
This PR adds all the tests from [WATs](https://github.com/cloudfoundry/wats) to CATs. This is intended as the first migration step to merge the two suites together. We have not removed the duplication in tests between the two suites, that will happen as part of subsequent cross-team work between the Garden Windows team and RelInt.

All tests from WATs are now in the `wats` package and the configurations for running these tests are now documented in the `README.md`. We have also added required fixture apps to the `assets` directory.


ishustava:
Hey @sunjayBhatia,

Thanks for the PR! We've run it, and it looks good on our end.

We'd like to make a few suggestions for the configuration parameters, though:
1. Could windows tests use existing parameter `credhub_mode` rather than introducing its own `windows_credhub_mode`? We think that most users of CATs would want those to be the same. 
1. We think it would be helpful for boolean parameters to have a verb to indicate that the value is either `true` or `false`. For example, we suggest renaming `windows_test_task` and `windows_context_path` to `use_windows_test_task` and `use_windows_context_path` respectively.
1. We liked that `windows_stack` mentions a default value in the README. We think it would be useful to mention defaults for other parameters, so that it's clear which ones are required.

We're happy to make this change if you prefer. We want to make sure that you are aware, so that your pipelines are updated accordingly. Let us know what you think. 

Thanks!
Iryna & Steve

sunjayBhatia:
@ishustava i think that sounds like a plan! we are not using this branch in our pipelines anywhere, we are still using WATs so feel free to change any config names if you like, it will not break us or anyone else

PR #273, opened 2018-03-09
Refactor Credhub service binding tests

benmoss:
- Extract helper methods to simplify
- Remove unnecessary test cleanup, the suite will ensure apps are torn
down
- Create the broker as a --space-scoped broker, no need to run as admin
- Only push the broker once, each other test can share it
- Combine some `It` blocks to avoid having to re-run all the test setup

### DIFF
--- a/credhub/helpers.go
+++ b/credhub/helpers.go
@@ -1,0 +1,168 @@
+package credhub
+
+import (
+	"io/ioutil"
+	"os"
+	"path"
+	"strings"
+
+	archive_helpers "code.cloudfoundry.org/archiver/extractor/test_helper"
+	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
+	"github.com/cloudfoundry-incubator/cf-test-helpers/workflowhelpers"
+	"github.com/cloudfoundry/cf-acceptance-tests/helpers/app_helpers"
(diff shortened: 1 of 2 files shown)


dsabeti:
@benmoss, cool, this mostly sounds good. One question: is there any risk for test pollution, given that the broker is pushed only once and shared by all the tests?

sunjayBhatia:
@dsabeti @benmoss this is a little tongue in cheek as Windows tests were just added to CATs, but is it worth getting these refactors over to the [`windows` package](https://github.com/cloudfoundry/cf-acceptance-tests/blob/73c1ff3d8cf9514a987b18e27f860f2728976319/windows/credhub_assist.go) as well or just to wait until all the tests are merged and duplication is removed?

benmoss:
@dsabeti Yeah, I worried about this too but I am pretty sure it's okay. The broker just sits there and writes credentials to CredHub using the apps instance GUIDs, so it's ok to handle multiple apps talking to it.

@sunjayBhatia hmm, I hadn't noticed that these got copied over as Windows tests.. that's a little unfortunate, but I'm sure we can try to de-duplicate lots of it. Lemme take a look and see how easy it can be.

benmoss:
I'm going to close this until we can figure out a way to reconcile the Windows ones and the Linux ones

PR #276, opened 2018-03-16
Change to check the variables with cf logs

andrew-edgar:
Currently the code uses the output from the cf start command.

it is definitely possible that the CLI may be slow connecting to doppler and thus misses these messages.

This causes a failure when it really should be a success as for this test the purpose is to test the variables are set not that you can see them on the cf cli start command.

This change therefore will use cf logs to retrieve the logs from the start and will correctly check if the variables are set.


dsabeti:
This change look good, and could potentially avoid some flakiness in the tests. I've prioritized it to get merged.

PR #278, opened 2018-03-22
Vendor in proxy test app

ameowlia:
[#155527932]

Signed-off-by: Amelia Downs <adowns@pivotal.io>

The merge for the following PR was reverted because we broke CATS with how packages were referenced in the proxy test app. 

To fix this we vendored the proxy app and are now re-submitting this PR to update running-security-group and container networking test to be one


ameowlia:
Oops, forgot to rebase in master since you reverted the last commit. Will fix and open another PR...again...

PR #280, opened 2018-03-27
Task CAT for compatibility with ASGs

chinangela:
This commit should be able to be merged on top of the other PR we made (vendor-proxy).

This PR is to add a new CAT to test that ASGs work with tasks. This test coverage was previously missing.

Please let me know if there are any questions or if there are any issues with merging.

[#155800937]

Thanks!
Angela


edespino:
We are using go1.10 and encountered the following error:

```
security_groups/running_security_groups.go:298: Sprintf format %d has arg privatePort of wrong type string
```

PR #282, opened 2018-03-30
Support only one backend (diego)

ishustava:
DEAs have been EOL'd, and CATs can drop its support for multiple backends and instead support just Diego. 

This PR makes the following changes:

* Remove 'backend' property from the config
* Remove tests that only run against the DEA backend
* Remove skip messages based on the backend property
* No longer call 'helpers.SetBackend' and collapse
  'cf push --no-start' with 'cf start' wherever possible.
* Ensure README no longer mentions backend requirements.


dsabeti:
🎉

charleshansen:
I was just about to do this myself as a CAPI thing until I was shown this PR. Looks like RelInt has it handled.

PR #290, opened 2018-04-19
Allow HTTP redirect in task overlay network tests

henrytk:
Platforms serving apps via HTTPS may use a redirect to promote from
HTTP. For such platforms using `curl` without `-L` produces an empty
response and consequent test failure.


henrytk:
Hi @heyjcollins,

We run on AWS. We have an ELB and HAProxy in front of GoRouter. The ELB has an HTTP listener and forwards to HAProxy. HAProxy [redirects HTTP to HTTPS](https://github.com/alphagov/paas-haproxy-release/blob/master/jobs/haproxy/templates/haproxy.conf.erb#L58-L65).

PR #291, opened 2018-04-23
Fix `GET /sigterm` endpoint for Dora

MirahImage:
- Implementation used to rely on `man` being installed
  - It is not installed on default CF buildpack
- Alternative implementation uses Ruby `Signal.list`

Pair with @blgm


blgm:
Thank you @njbennett.  I don't know why it didn't work for you.  Maybe it's to do with the Ruby version or OS.  For us it went from not working to working.  Thank you for merging.

PR #299, opened 2018-05-11
Use go app for windows metrics tests

natalieparellano:
This app was based off of the loggregator load generator
Sinatra app.

Our intention is to have the windows loggregator tests match the
linux tests as much as possible. Hopefully this will remove some of the
flakiness.

Maybe in the future the linux loggregator tests could utilize this app as well.


dsabeti:
Thanks for this @natalieparellano. I'll prioritize it to get merged. 

In the meantime, I wonder if @johannasmith has any opinions about using this in the normal loggregator tests. Also, I've recently been thinking trying to unify the Windows tests with some of the related Linux-oriented tests -- it would be nice if you could merely specify a Stack to target, and the tests would otherwise operate the same way. (cc @staylor14)

johannasmith:
I haven't interacted with the sinatra app, but we would be interested in running the upgraded windows tests.

PR #302, opened 2018-05-24
Skip security-group c2c tests when not using container-networking.

andreas-kupries:
Running the CATs in a setup without container-networking broke the security-group test case `[It] correctly configures asgs and c2c policy independent of each other`. Inspection showed that while the failed test depends on c2c, its section is not conditional on it, causing execution even when c2c is not present/disabled.

This commit fixes the issue in the `cf.135` branch where it was found.


ishustava:
Hey @andreas-kupries,

This bug is fixed on `cf1.36` branch of CATs. 

Given that cf-deployment `v1.35` is marked as pre-release, we recommend that you upgrade cf-deployment and use `cf1.36` branch of CATs.

We're going to close this PR for now, but feel free to re-open it if you see any issues.

Thanks,
Iryna & Yash
CF Release Integration Team

Issue #303, opened 2018-05-29
apps/loggregator.go test fails with log_cache_enabled 


keymon:
The following test fails when using log-cache as it was enabled in https://github.com/cloudfoundry/cf-acceptance-tests/commit/923a0fe9335bf0a1e75286801e952f248491cd38 by @MasslessParticle and @bradylove 

```
• Failure [159.719 seconds]
[apps]
/tmp/build/0d76c07a/src/github.com/cloudfoundry/cf-acceptance-tests/cats_suite_helpers/cats_suite_helpers.go:31
  loggregator
  /tmp/build/0d76c07a/src/github.com/cloudfoundry/cf-acceptance-tests/cats_suite_helpers/cats_suite_helpers.go:37
    cf logs
    /tmp/build/0d76c07a/src/github.com/cloudfoundry/cf-acceptance-tests/apps/loggregator.go:55
      exercises basic loggregator behavior [It]
      /tmp/build/0d76c07a/src/github.com/cloudfoundry/cf-acceptance-tests/apps/loggregator.go:69

      Timed out after 120.000s.
      Got stuck at:
             2018-05-29T15:53:41.66+0000 [CELL/0] OUT Starting health monitoring of container
             2018-05-29T15:53:42.69+0000 [APP/PROC/WEB/0] ERR [2018-05-29 15:53:42] INFO  WEBrick 1.3.1
             2018-05-29T15:53:42.69+0000 [APP/PROC/WEB/0] ERR [2018-05-29 15:53:42] INFO  ruby 2.4.3 (2017-12-14) [x86_64-linux]
             2018-05-29T15:53:42.69+0000 [APP/PROC/WEB/0] ERR [2018-05-29 15:53:42] INFO  WEBrick::HTTPServer#start: pid=7 port=8080
             2018-05-29T15:53:43.39+0000 [CELL/1] OUT Starting health monitoring of container
             2018-05-29T15:53:43.76+0000 [CATS-5-APP-1372e34d96177fc4/0] GAUGE cpu:0.000000 percentage disk:51441664.000000 bytes disk_quota:1073741824.000000 bytes memory:40960.000000 bytes memory_quota:268435456.000000 bytes
             2018-05-29T15:53:43.77+0000 [CELL/0] OUT Container became healthy
             2018-05-29T15:53:44.95+0000 [APP/PROC/WEB/1] ERR [2018-05-29 15:53:44] INFO  WEBrick 1.3.1
[...]

PR #304, opened 2018-05-29
cf tail doesn't expect header

johannasmith:
If running with tail, header will not be emitted
- CATs will never run as a terminal

[#157710932]

Signed-off-by: Todd Persen <tpersen@pivotal.io>

/cc @toddboom

### DIFF
--- a/apps/loggregator.go
+++ b/apps/loggregator.go
@@ -67,7 +67,12 @@ var _ = AppsDescribe("loggregator", func() {
 		})
 
 		It("exercises basic loggregator behavior", func() {
-			Eventually(logs, (Config.DefaultTimeoutDuration() + time.Minute)).Should(Say("(Connected, tailing|Retrieving) logs for app"))
+			if Config.GetUseLogCache() {
+				// log cache cli will not emit header unless being run in terminal
+				Consistently(logs, (Config.DefaultTimeoutDuration() + time.Minute)).ShouldNot(Say("(Connected, tailing|Retrieving) logs for app"))
+			} else {
+				Eventually(logs, (Config.DefaultTimeoutDuration() + time.Minute)).Should(Say("(Connected, tailing|Retrieving) logs for app"))
+			}
 
 			Eventually(func() string {
 				return helpers.CurlApp(Config, appName, fmt.Sprintf("/log/sleep/%d", hundredthOfOneSecond))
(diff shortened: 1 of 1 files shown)


dsabeti:
@johannasmith, I'm guessing this addresses this issue: https://github.com/cloudfoundry/cf-acceptance-tests/issues/303 ?

johannasmith:
@dsabeti correct.

apps/loggregator.go
```diff
@@ -67,7 +67,12 @@ var _ = AppsDescribe("loggregator", func() {
 		})
 
 		It("exercises basic loggregator behavior", func() {
-			Eventually(logs, (Config.DefaultTimeoutDuration() + time.Minute)).Should(Say("(Connected, tailing|Retrieving) logs for app"))
+			if Config.GetUseLogCache() {
+				// log cache cli will not emit header unless being run in terminal
+				Consistently(logs, (Config.DefaultTimeoutDuration() + time.Minute)).ShouldNot(Say("(Connected, tailing|Retrieving) logs for app"))
```
dsabeti: Hey @johannasmith. I think we could probably do without this assertion. Since it's a negative assertion inside of a `Consistently`, the assertion is almost always going to run through each poll for the entire timeout; it will take `DefaultTimeoutDuration` plus 1 minute to run. Since there's [a good assertion below](https://github.com/cloudfoundry/cf-acceptance-tests/pull/304/files#diff-9043d6eb59a3fedf9989ec85b9b04cd5R81), maybe just have the guard clause do nothing when Log Cache is being used:
```
if !Config.GetUseLogCache() {
  Eventually(logs, ...)
}
```

toddboom: @dsabeti Sounds good. We pushed the change you suggested. We're re-running the tests in our CI and will ping you when they go green.

johannasmith: @dsabeti they passed. good to pull in this change

toddboom:
@dsabeti do you want us to merge this in ourselves, or would you prefer to be the button clicker?

dsabeti:
@toddboom, no, it's prioritized in the backlog. An engineer from the RelInt team will get to it soon and merge it in.

staylor14:
Since the name of the test is "cf logs" "exercises basic loggregator behavior", it's probably ok for the test to alter its behavior based on the configuration, but I'm skeptical about a test that doesn't specifically call out what behaviors it expects.  I'm dropping this comment in here so that we (the Release Integration team) can reason about these decisions as we look at how CATs are structured generally.

staylor14:
I have to learn how github works, apparently.  I meant to close the comment, not "close AND comment".  blargh.

PR #307, opened 2018-06-19
add internal domain as a config param

ameowlia:
Hi RelInt, 

Container Networking no longer assumes that the internal domain `apps.internal` is present and instead requires the user to configure their own internal domain. This PR reflects those changes.

<hr>

- for service discovery
- no longer assume that apps.internal is present, since it is no longer
seeded by capi or hardcoded in the bosh dns adapter
- users must now BYO internal domain

[#158296930]

Signed-off-by: Amelia Downs <adowns@pivotal.io>


ishustava:
To capture the discussion with @christianang & @tylerschultz:

We are going to remove `internal_domain` config parameter in an effort to reduce the number of configuration parameters needed to run CATs. We'll instead use hard-coded default internal domain `apps.internal` from cf-networking release and wait until somebody requests to configure this to a different value.

PR #312, opened 2018-07-11
Increase timeout in loggregator firehose test.

keymon:
We've seen this test failing intermittently. Increasing the timeout here
seems to resolve this. This also changes it to match the container
metrics test below which also uses 2*config.DefaultTimeoutDuration which
was increased in story 1995884 to fix a similar issue.

In my testing, I often saw this failing against an AWS deployment,
whereas with this increased timeout, it didn't file after 20 rounds
using `ginkgo -untilItFails`. With some tracing added, I often saw it
taking a little over a minute for the messages to appear.

/cc @alext original author.

### DIFF
--- a/apps/loggregator.go
+++ b/apps/loggregator.go
@@ -2,6 +2,7 @@ package apps
 
 import (
 	"fmt"
+
 	. "github.com/cloudfoundry/cf-acceptance-tests/cats_suite_helpers"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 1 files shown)


ishustava:
Hey @keymon,

Thanks for this PR.

We've seen quite a bit of flakiness as well with this test in the past, but since [this commit](https://github.com/cloudfoundry/cf-acceptance-tests/commit/a3a23309c12e4d9c2a80fc1d314c3f46d5516356) its flake rate is zero. So, we would like to understand the source of flakiness you're seeing a bit better before we increase the timeout.

We found that running this test in isolation with `-untilItFails` flag doesn't give us confidence that it's stable because the flakiness we were seeing only happened when all tests ran. When all tests run, the number of messages available on the firehose `msgChan` increases significantly. I'm curios if you run this test as is (without the increased timeout) with `-untilItFails`, can you reproduce the flakiness? 

It would also be good to know the following:
1. How are you deploying CF? Is this a bosh-lite environment?
1. Which IaaS are you using?
1. How many ginkgo nodes are you using?

keymon:
1. How are you deploying CF? Is this a bosh-lite environment?

We have a full deployment on AWS based on cf-deployment with some extra VMs
for brokers and logsearch

   1. Which IaaS are you using?

Aws

   1. How many ginkgo nodes are you using?

We run 5 nodes. We do not run it as a errand but from concourse as a job,
in the same box, although I guess that is irrelevant.

https://github.com/alphagov/paas-cf/blob/master/platform-tests/upstream/run_acceptance_tests.sh

ishustava:
Ok, thanks for the info.

A few more follow-up questions:
1. Which version of cf-acceptance-tests and CF were you seeing the failure with?
1. When you ran `ginkgo -untilItFails` and it passed for 20 rounds, did you run the whole suite or just that one test in isolation?

alext:
Hi, I'm a colleague of @keymon.

> Which version of cf-acceptance-tests and CF were you seeing the failure with?

This was with 1.38. I can re-test against 2.6 if you think that's necessary.

> When you ran ginkgo -untilItFails and it passed for 20 rounds, did you run the whole suite or just that one test in isolation?

This was with the one test focussed. Without this change it would fail in around 3 runs with the test focussed.

alext:
Hi @ishustava, I've re-run the tests using the `cf2.6` branch, and can no longer reproduce the issue. I think this can therefore be closed.

thanks for the response.

PR #313, opened 2018-07-12
Bumped credhub cli

TisVictress:
In the newest version of credhub, the `overwrite` option is not allowed when setting credentials. This bump updates the broker to use a compatible client.

* Updating import path of credhub
* Update service broker

@mdelillo 
Signed-off-by: Mark DeLillo <mdelillo@pivotal.io>


njbennett:
Hey friends, thanks for this PR. We have one request. 

It looks like you've got a bunch of files vendored in that might not need to be. We think that adding
```
[prune]
  go-tests = true
  unused-packages = true
```
to your `Gopkg.toml` file will do the trick. We confirmed this removes a bunch of files, but we haven't tested it and we don't know what all the tests that would need to be run to check that the broker is still doing everything you want it to do.

Can you update the `toml` file and run whatever tests you'd want to run, and then update this PR? Alternatively if all the testing you would do is run CATs yourself we can update the file and then get it into our pipeline.

mdelillo:
@njbennett updated Gopkg per your suggestion, credhub tests are still passing.

njbennett:
Thanks!

PR #315, opened 2018-07-18
Add protocol to service discovery app url.

jmcarp:
We run a proxy in front of the gorouter that redirects http to https, so this test fails without adding the protocol to the url.


jmcarp:
Tagging recent committers: @ishustava @njbennett @heyjcollins. I think this should be quick to review, and it would be helpful for us to get it merged when you all have time.

PR #318, opened 2018-07-26
App env vars are different when unproxied communication is disabled

jvshahid:

As part of [#157675252](https://www.pivotaltracker.com/story/show/157675252), operators can now configure Diego to require that all traffic to app containers go through the Envoy proxy.  In that scenario  `CF_INSTANCE_PORTS` & `CF_INSTANCE_PORT` environment variable will no longer be set.  Furthermore, `CF_INSTANCE_ADDR` won't have the 'external' field set. This commit introduces the following changes:

- New parameter 'disallow_unproxied_app_traffic'.
- Change the relevant tests to use the previously mentioned parameter.

https://www.pivotaltracker.com/story/show/157675252. Also chatted with @njbennett @ytulsiani about the change, so they should have some context as well.

cloudfoundry/cf-deployment@030626013a80c9bc78f33561add4f3cc4e2485fd. Note that Running CATs with the new flag `disallow_unproxied_app_traffic` set to `true` requires using the latest [diego-release](https://github.com/cloudfoundry/diego-release/tree/develop) with `containers.proxy.enable_unproxied_port_mappings` set to `true` on the `Rep` job.

- [ ] introduces a new test --- are you sure everyone should be running this test?
- [X] requires an update to a CATs integration-config

- [X] YES
- [ ] N/A

Add a new config parameter to change the behavior of the tests when unproxied port mapping is disallowed in Diego.

This change shouldn't affect the runtime of CATs. It only replaces existing assertions with more appropriate assertions if `disallow_unproxied_app_traffic` was set to `true`.

This is currently blocking Diego from merging this feature and testing it in CI.

@cloudfoundry/cf-diego

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,8 @@ include_capi_no_bridge
 * `private_docker_registry_password`: Password to access the private docker repository. [See below](#private-docker)
 * `unallocated_ip_for_security_group`: An unused IP address in the private network used by CF. Defaults to 10.0.244.255. [See below](#container-networking-and-application-security-groups)
 
+* `disallow_unproxied_app_traffic`: Set this to `true` if Diego was configured to disallow unproxied port mappings, i.e. if `containers.proxy.enable_unproxied_port_mappings` is set to `false`.  Note that this also requires using the [cf-syslog-skip-cert-verify](https://github.com/cloudfoundry/cf-deployment/blob/master/operations/cf-syslog-skip-cert-verify.yml).
+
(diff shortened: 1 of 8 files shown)


jvshahid:
We decided to move on with merging the changes in diego-release instead of blocking on the CATs changes being merged in. Updated the description to reflect that, previously the description mentioned that the changes were on a branch in diego-release which isn't true anymore.

davewalter:
The story associated with this PR was accepted back in 2018. Going to close this PR without merging.

PR #333, opened 2018-09-04
Allow security groups to pass on NSX-T

cunnie:
This is the second commit to enable CF foundations using the NSX-T
container plugin (NCP) to pass the security groups test.

Specifically, it accommodates NSX-T's behavior of notifying the
container that the host is unreachable (by proactively sending ICMP host
unreachable packets) rather than the "normal" behavior of simply
allowing connection attempts to time out.

<https://en.wikipedia.org/wiki/Internet_Control_Message_Protocol#Destination_unreachable>

Note that NSX-T only manifests this behavior for IP pools which it
manages, not for the internet at large.

Signed-off-by: Rowan Jacobs <rojacobs@pivotal.io>

Yes

Accommodate the subtle differences that NSX-T exhibits when blocking traffic (versus Silk)

See #331

PAS 2.3.0-beta2

- [ ] YES
- [X] N/A

N/A

See #331

None 

/cc @rowanjacobs @gregoehmen

### DIFF
--- a/tasks/task.go
+++ b/tasks/task.go
@@ -349,7 +349,7 @@ exit 1`
 					appLogs := logs.Tail(Config.GetUseLogCache(), appName).Wait()
 					Expect(appLogs).To(Exit(0))
 					return string(appLogs.Out.Contents())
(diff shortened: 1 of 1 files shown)


rowanjacobs:
(Title and description is misleading; this is actually for the v3 tasks test.)

PR #336, opened 2018-09-08
Fix text matching to be similar to the non-windows instance reporting tests

jpalermo:

We're seeing the windows instance reporting specs fail with the following error. Not sure what caused the breakage, but making the test more similar to the `apps/instance_reporting.go` test has resolved the problem for us. Possibly related to a switch to go 1.10 in our pipelines.

```
[windows] Getting instance information scaling memory beyond the available amount [It] fails with insufficient resources 
/tmp/build/33ac16d1/go/src/github.com/cloudfoundry/cf-acceptance-tests/windows/instance_reporting.go:55

  Got stuck at:
      Showing health and status for app CATS-1-APP-f75554cb5e07d8ec in org CATS-1-ORG-8621f09246baa562 / space CATS-1-SPACE-ede9ebe9035b6d2d as CATS-1-USER-f87d069a567b368a...
      
      name:              CATS-1-APP-f75554cb5e07d8ec
      requested state:   started
      routes:            cats-1-app-f75554cb5e07d8ec.apps.swamp-binder.gcp.releng.cf-app.com
      last uploaded:     Fri 07 Sep 23:33:21 UTC 2018
      stack:             windows2016
      buildpacks:        binary_buildpack
      
      type:           web
      instances:      0/1
      memory usage:   921600M
      
      There are no running instances of this process.
      
  Waiting for:
      insufficient resources

  /tmp/build/33ac16d1/go/src/github.com/cloudfoundry/cf-acceptance-tests/windows/instance_reporting.go:58
```

Custom deployment with the latest version of the components I believe.

- [ ] YES
- [X] N/A

Just a refactor of the matching string, probably not worth mentioning in release notes.

N/A

### DIFF
--- a/windows/instance_reporting.go
+++ b/windows/instance_reporting.go
@@ -39,23 +39,23 @@ var _ = WindowsDescribe("Getting instance information", func() {
 		Eventually(cf.Cf("delete", appName, "-f")).Should(Exit(0))
 	})
 
-	Context("scaling memory beyond the available amount", func() {
+	Context("scaling memory", func() {
 		BeforeEach(func() {
 			setTotalMemoryLimit(RUNAWAY_QUOTA_MEM_LIMIT)
-
-			scale := cf.Cf("scale", appName, "-m", EXCEED_CELL_MEMORY, "-f")
-			Eventually(scale, Config.CfPushTimeoutDuration()).Should(Say("insufficient resources|down"))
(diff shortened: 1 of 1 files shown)


mcwumbly:
@heyjcollins - is there anything we can do to help move this PR along?

PR #339, opened 2018-09-12
Use CF CLI to create and delete service bindings

henryaj:
Update service binding tests to use the CF CLI instead of `cf curl`. This new test is designed to work with both the current wording and the updated wording introduced in https://github.com/cloudfoundry/cli/pull/1452.

https://www.pivotaltracker.com/story/show/159168130

- [ ] YES
- [x] N/A

no change

@georgi-lozev

### DIFF
--- a/services/service_instance_lifecycle.go
+++ b/services/service_instance_lifecycle.go
@@ -60,12 +60,12 @@ var _ = ServicesDescribe("Service Instance Lifecycle", func() {
 		}, Config.AsyncServiceOperationTimeoutDuration(), ASYNC_OPERATION_POLL_INTERVAL).Should(Say("not found"))
 	}
 
-	waitForAsyncOperationToComplete := func(broker ServiceBroker, instanceName string) {
+	waitForAsyncOperationToCompleteAndSay := func(broker ServiceBroker, instanceName, expectedText string) {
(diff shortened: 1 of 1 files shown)


georgi-lozev:
Keep in mind the minimum required version of the CLI in the environment where you're going to execute the tests is [v6.39.0](https://github.com/cloudfoundry/cli/releases/tag/v6.39.0).

henryaj:
Ping @heyjcollins @ishustava - let us know if you need anything else from us on this. :)

heyjcollins:
Hey @henryaj - if this requires v6.39.0 of the CLI, we're actually blocked from moving forward untill the CLI team can make an update to support another CAT that started failing. We're pinned to an earlier version 'till that's resolved. I can let them know you're blocked so up the anti.

nmaslarski:
Thank you for the attention!
Yes, this does requires 6.39.0.
It's blocking only this PR, you can ping them if you like.
Thanks.

henryaj:
Hi @heyjcollins - just wanted to ping you on this. Did the CLI team get round to making that fix?

mattmcneeney:
[v6.40.0](https://github.com/cloudfoundry/cli/releases) is the latest release, so this should be good to go!

PR #343, opened 2018-09-19
Attempt to reduce flakiness around broker healthchecks

ljfranklin:
We intermittently see CATs which push a service broker fail with:

```
   2018-09-19T01:26:29.43+0000 [APP/PROC/WEB/0] ERR 10.255.21.109 - - [19/Sep/2018:01:26:29 +0000] "GET /v2/catalog HTTP/1.1" 200 2306 1.5427
   2018-09-19T01:26:30.09+0000 [HEALTH/0] ERR Failed to make HTTP request to '/v2/catalog' on port 8080: timed out after 1.00 seconds
   2018-09-19T01:26:30.09+0000 [CELL/0] OUT Container became unhealthy
   2018-09-19T01:26:30.09+0000 [APP/PROC/WEB/0] ERR [2018-09-19 01:26:30] FATAL SignalException: SIGTERM
```

This PR bumps the healthcheck timeout from 1 second to 10 seconds. We
had to switch to `cf v3-set-health-check` as it didn't look like the v2
command supported a timeout

- [ ] YES
- [x] N/A

_Something brief that conveys the change and is written with the component author audience in mind._

### DIFF
--- a/helpers/services/broker.go
+++ b/helpers/services/broker.go
@@ -136,7 +136,7 @@ func (b ServiceBroker) Push(config cats_config.CatsConfig) {
 		"-p", b.Path,
 		"-d", config.GetAppsDomain(),
 	).Wait(Config.BrokerStartTimeoutDuration())).To(Exit(0))
-	Expect(cf.Cf("set-health-check", b.Name, "http", "--endpoint", "/v2/catalog").Wait(Config.BrokerStartTimeoutDuration())).To(Exit(0))
+	Expect(cf.Cf("v3-set-health-check", b.Name, "http", "--endpoint", "/v2/catalog", "--invocation-timeout", "10").Wait(Config.BrokerStartTimeoutDuration())).To(Exit(0))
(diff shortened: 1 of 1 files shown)


Changdrew:
With the advent of the v7 CLI, supporting a v3 cli command isn't a viable long term solution.

ljfranklin:
@Changdrew the test as currently written fails intermittently as the healthcheck timeout is too low. How would you like to address this other than using `v3-set-healthcheck`?

ljfranklin:
Going to re-open so this issue doesn't get lost

ishustava:
@ljfranklin,

Per our in person conversation, I'm going to close it for now. Please feel free to re-open it (again) if you see this test becoming flaky again, we'd like to take a look at it. 

Our thinking is as follows:
- As @Changdrew mentioned above, since this PR adds a `v3` command, it will put CATs in a weird position. As far as we know CF CLI plans to remove their experimental `v3` commands before CLI v7, which will use v3 API out of the box, will become GA (cc @abbyachau @tjvman)
- We can't think of another way to change the invocation timeout other than your suggestion of calling `v3` API directly rather than using CF CLI. 

Hope this makes sense.

PR #346, opened 2018-10-09
Add internetless test suite

genevieve:

Adding a test suite that verifies a deployment does not allow external traffic from app containers.

Slack: https://pivotal.slack.com/archives/C3LUD2L04/p1538515665000100
Story: https://www.pivotaltracker.com/story/show/160211045

Not cf-deployment...

Technically, there is a new config option for the new test, so if the user wants to run that test, they would need to update their CATs integration-config but otherwise, it defaults to false.

- [x] YES
- [ ] N/A

See slack conversation.

Adds a test that verifies a deployment does not allow external traffic from app containers.

0s if not enabled.

@acrmp @mcwumbly

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ cat > integration_config.json <<EOF
   "include_detect": true,
   "include_docker": false,
   "include_internet_dependent": false,
+  "include_internetless": false,
   "include_isolation_segments": false,
   "include_private_docker_registry": false,
   "include_route_services": false,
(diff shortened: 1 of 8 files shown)


acrmp:
This also needs the import for the new test to be added to `cats_suite_test`.
https://github.com/genevieve/cf-acceptance-tests/blob/f85eb542b1540a41f7fb8eb68276374427667693/cats_suite_test.go#L21

PR #354, opened 2019-01-25
Improve error reporting when API redirects multiple times

petergtz:

yes

'can copy package bits to another app and download the package' in `v3/package.go` fails with a confusing unzip error, when api redirects multiple times, because it tries to unzip the body of a redirect. This change stops on the second redirect with an error message emulating what the cf CLI would do.

I've started a [conversation on the cf CLI Slack](https://cloudfoundry.slack.com/archives/C032824SM/p1548426413059900) and it seems like it is not on purpose that the CLI doesn't follow more than 1 redirect. Once it's changed in the cf CLI, it would be preferable in the code that I've refactored here, to actually just follow all redirects just as the CLI does to be in parity. Then the improved error reporting introduced here might not be needed anymore, unless of course you decided to still stop redirecting after 10 redirects or so. Would still be nicer to get an error about redirecting instead of a failed unzip.

master

- [ ] YES
- [x] N/A

N/A

> Improve error message when API redirects multiple times.

About the same.

@idev4u

### DIFF
--- a/helpers/download/download.go
+++ b/helpers/download/download.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"regexp"
+	"strconv"
 
 	"github.com/cloudfoundry-incubator/cf-test-helpers/helpers"
 	"github.com/cloudfoundry/cf-acceptance-tests/helpers/config"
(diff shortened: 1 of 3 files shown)


cdutra:
Hi @petergtz, thanks for the PR!

It makes sense for the cf cli and curl in our tests to follow multiple redirects. However, I have a few suggestions to the code, to avoid more repetition and at least implement a better multiple redirect logic with a loop or recursion with maybe a predefined depth.

Thanks!

petergtz:
Hi @cdutra,

Makes sense. Could you please have a look at it with my latest commit? In addition to moving everything into a loop to allow for an arbitrary number of redirects, I set the max number of redirects to 10, which is what Go's default http client does.

Additionally, the CLI team responded that the fact that they only do 1 redirect is for unknown legacy reasons. They have a [story to fix it](https://www.pivotaltracker.com/n/projects/892938/stories/163533146/expand). For the Bits-Service having this limitation of 1 redirect is a real pain. Therefore, I'd be happy if we could change it in CATs even before this is fixed in the CLI. What do you think?

Thanks,
Peter

cdutra:
Hi @petergtz,

It looks good, thank you so much for the quick response! We appreciate that.
We'll merge the PR.

Thanks!
Chris

PR #360, opened 2019-02-28
Make cats run with Go 1.12

akshaymankar:

Make CATS run with Go 1.12
GOCACHE is now required but it can be calculated using 
HOME or XDG_CACHE_HOME, so we are passing it from environment

https://golang.org/doc/go1.12#gocache

SUSE CF 2.14.5

- [ ] YES
- [x] N/A

No new tests introduced.

Doesn't need to be described in release notes.

0s

@alex-slynko


cdutra:
Hi @akshaymankar, thanks for the PR!

We got another [PR](https://github.com/cloudfoundry/cf-acceptance-tests/pull/362) solving the same issue. We believe using `os.Environ()` is better because this will avoid future changes if later versions of Golang require other environment variables to be set. Feel free to reopen the PR if you have any concerns.

Thanks!

PR #367, opened 2019-04-10
Remove `include_assisted_credhub` from ex. config

mcwumbly:
It doesn't appear to be a real thing.


vitreuz:
It certainly is not a thing

PR #374, opened 2019-04-26
fix container metrics regex to accomodate apps with less memory

cwlbraa:
Co-authored-by: Connor Braa <cbraa@pivotal.io>
Co-authored-by: Renee Chu <rchu@pivotal.io>

sometimes the test app here uses an amount of memory measured in K, not M or G. This causes the test to fail erroneously.

- [ ] YES
- [x] N/A

fix container metrics regex to accomodate apps with less memory

@reneighbor @cdutra


cwlbraa:
renee publicized her cloudfoundry membership

PR #385, opened 2019-08-23
Volume services test uses a cf-pushed nfs server

paulcwarren:
[#167932835](https://www.pivotaltracker.com/story/show/167932835)

> Yes

> Having the ability to run volume-services suite without any external nfs server running. Instead the suite 'creates' a nfs server to run tests against.

> We would like it to be easier to run the volume services suite

> https://github.com/cloudfoundry/cf-deployment/commits/29bd9ac01754398587623435403437da894f6684

- [ X ] changes an existing test
- [ X ] requires an update to a CATs integration-config

> We removed an unused configuration property `volume_service_create_config`

- [ X ] YES
- [ ] N/A

n/a

Volume services test now uses a cf-pushed "in-memory" nfs server

~1 minute

- [ X ] **Urgent** - unblocks current or future work

cc/ @DennisDenuto @julian-hj


paulcwarren:
Folks,
We updated the PR as requested.  Please let us know if there are any other issues/comments.
Thx

PR #386, opened 2019-09-17
Run tests faster with mod=vendor

akshaymankar:

Yes

As all of the dependencies are vendored, we should leverage that. This saves about a minute in our (Eirini team's) CI.

N/A

SCF + Eirini

- [ ] YES
- [x] N/A

N/A

cf-acceptance-tests will leverage vendored modules using `-mod=vendor`. This means go versions less than 1.11 will not be supported.

This change cuts down 55 seconds.

@alex-slynko


weymanf:
I had to revert this change until we figure out how to run this in our pipelines. Looks like it causes a lot of problems if a bazaar package is not installed.

PR #396, opened 2019-12-05
assert different "routing transparency" behavior for Istio Proxy

tcdowney:

Ya 👍 

Istio performs some URL normalization so it does not have _complete_ route transparency like the Gorouters do. For various reasons this is desired behavior on the Istio side and they will not be changing it. We think that based on the history of this CAT (https://github.com/cloudfoundry/gorouter/issues/46) that this behavior is fine to differ between Gorouter and Istio-Proxy.

This change does the following:
- adds a new config property, `UseKubernetesIstioProxy` to use as a flag
for asserting on behavior differences between Gorouter and Istio Proxy. This property defaults to `false` to maintain the current CATS behavior.
- modifies the routing transparency CATS to allow for Istio Proxy's URL
normalization

Story: https://www.pivotaltracker.com/story/show/169404143

We've run the existing Gorouter-specific tests against an environment using `release-candidate` of cf-deployment. We ran the Istio-Proxy specific CATS against an Eirini environment.

- [x] YES
- [ ] N/A

This is an existing test that applies differently in the Istio-Proxy/CF on k8s world. We are introducing change this to CATS so that we can continue to run CATS against environments that do not use Gorouters.

* assert different "routing transparency" behavior for Istio Proxy

It will not affect CATS runtime. Same amount of tests, just different assertions.

@ndhanushkodi @mike1808 @rosenhouse @rodolfo2488

### DIFF
--- a/apps/routing_transparency.go
+++ b/apps/routing_transparency.go
@@ -13,12 +13,11 @@ import (
 	"github.com/cloudfoundry/cf-acceptance-tests/helpers/app_helpers"
 	"github.com/cloudfoundry/cf-acceptance-tests/helpers/assets"
 	"github.com/cloudfoundry/cf-acceptance-tests/helpers/random_name"
+	"github.com/cloudfoundry/cf-acceptance-tests/helpers/skip_messages"
 )
 
 var _ = AppsDescribe("Routing Transparency", func() {
 	var appName string
-
-	SkipOnK8s("See https://github.com/cloudfoundry/cf-acceptance-tests/pull/396")
 
 	BeforeEach(func() {
 		appName = random_name.CATSRandomName("APP")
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -144,6 +144,7 @@ include_capi_no_bridge
 * `include_zipkin`: Flag to include tests for Zipkin tracing. `include_routing` must also be set for tests to run. CF must be deployed with `router.tracing.enable_zipkin` set for tests to pass.
 * `use_http`: Set to true if you would like CF Acceptance Tests to use HTTP when making api and application requests. (default is HTTPS)
 * `use_log_cache`: Set to false if you don't want CF Acceptance Tests to use Log Cache for reading application logs. (default is true)
+* `use_kubernetes_istio_proxy`: Set to true if running against a "Cloud Foundry for Kubernetes" environment using [cf-k8s-networking](https://github.com/cloudfoundry/cf-k8s-networking) with an Istio Proxy for ingress routing. (default is `false`)
```
tcdowney: Open to suggestions for this config flag name. We left it a bit generic because we may need to tweak tests around other Gorouter-specific behavior in the future.

acosta11:
Hello Networking team,

Firstly, sorry about the delay here given the marked urgency of the request. We've had a lot of trouble getting the team together for technical retro recently.

The RelInt team discussed the proposed change in technical retro, and identified the consequences of pulling in a functionality change between bosh deployed go router and kubernetes deployed istio network resources as a risk we don't want to take on the main CATs branch right this moment.

The potential class of differences between bosh and kubernetes environments may warrant a more generic configuration convention or another mechanism altogether to sufficiently surface and enforce that distinction. With the history of config and complexity creep in CATs in mind, we want to be careful at this point and do the right thing.

But given the urgency of the concern, we don't want to block your local development and CI test feedback. We considered the potential workaround of skipping the failing tests in k8s CI by passing [a  parameter](https://github.com/cloudfoundry/cf-deployment-concourse-tasks/blob/eaaf086771f62315ca75a48ba1e8ebad272ad2a4/run-cats/task.yml#L45) through the run-cats concourse task that ginkgo interprets to skip test names matching a specified regular expression and then temporarily defining that istio-specific test in a separate script in CI in the meantime to bridge the coverage gap. The cost there would be not explicitly testing the istio routing transparency behavior in CATs to maintain the balance in the bosh world.  Apologies if this is something you've already considered, but would it work for the immediate need?

Thanks,
Andrew and the RelInt team

tcdowney:
@acosta11, yeah we currently have it skipped. We mostly just want to encode that we accept the way this works in Istio in the CAT before we lose context on it. Also as other component teams that run CATS adopt https://github.com/cloudfoundry/cf-k8s-networking we figured this would be a good temporary solution for them to get CATS passing without having to communicate that they need to skip "Routing Transparency" cats.

> With the history of config and complexity creep in CATs in mind, we want to be careful at this point and do the right thing.

Do you have the concept of of `temporary_` or `experimental_` feature flags? I feel like this is the type of thing that could easily be removed once you have the more holistic solution. Especially since CATS are versioned now and it is configured to retain the gorouter behavior by default.

rosenhouse:
Bump.  @acosta11 does your team have any thoughts on the suggestion made by @tcdowney above?  We wonder how you'll get CATS passing on the CF-on-K8s artifact.

apps/routing_transparency.go
```diff
@@ -33,17 +34,47 @@ var _ = AppsDescribe("Routing Transparency", func() {
...
+			Expect(curlResponse).To(ContainSubstring("/requesturi/%21%7E%5E%24%20%27%28%29"))
+			Expect(curlResponse).To(ContainSubstring("Query String is [foo=bar+baz%20bing]"))
+		})
+
+		It("transparently proxies both reserved characters and unsafe characters", func() {
+			curlResponse := helpers.CurlApp(Config, appName, "/requesturi/!~^'()$\"?!'()$#!'")
+			Expect(curlResponse).To(ContainSubstring("Request"))
+			Expect(curlResponse).To(ContainSubstring("/requesturi/!~^'()$\""))
+			Expect(curlResponse).To(ContainSubstring("Query String is [!'()$]"))
+		})
```
ericpromislow, line 78: It took me a while to see how the two separate blocks differ.  I would have preferred something like this:
```
var tildeRepresentation string
if Config.GetUseKubernetesIstioProxy() {
  tildeRepresentation = "~"
} else {
  tildeRepresentation = "%7E"
}
It("normalizes URLs with percent-encoded characters by decoding certain characters", func() {
	curlResponse := helpers.CurlApp(Config, appName, "/requesturi/%21%7E%5E%24%20%27%28%29?foo=bar+baz%20bing")
	Expect(curlResponse).To(ContainSubstring("Request"))
	Expect(curlResponse).To(ContainSubstring("/requesturi/%21"  + tildeRepresentation + "%5E%24%20%27%28%29"))
	Expect(curlResponse).To(ContainSubstring("Query String is [foo=bar+baz%20bing]"))
})
...
```

acosta11:
> Bump. @acosta11 does your team have any thoughts on the suggestion made by @tcdowney above? We wonder how you'll get CATS passing on the CF-on-K8s artifact.

General status update on CATs:
Ultimately, because of blockers to even a basic CATs run, we haven't yet been able to prioritize further CATs work of this type. 😞 

We tried running the full CATs suite when we finished the kpack cnb integration, but immediately encountered issues with cf buildpacks cli and api functionality. More detail is available at: https://www.pivotaltracker.com/story/show/171293818.

README.md
jamespollard8: Would it make sense to remove the Kubernetes reference in these? What would we call it if we'd enabled using Istio in cf-deployment?

jamespollard8:
Update: we're ready to work on merging this in. We don't need to block it on the other CATS work we're doing - as long as CATS continue to run fine in our cf-deployment pipelines.

README.md
tcdowney: Probably `use_istio_ingress` or `istio_ingress_enabled` or something. 🤔  I agree we could get rid of the `kubernetes` part of the name and that the "istio" part is more important. I imagine if we changed ingress implementations that this behavior would change as well.

jamespollard8: `use_istio_ingress` sounds good to me 👍

tcdowney:
@jamespollard8 @ericpromislow 

I rebased this branch and made the change to switch from `use_kubernetes_istio_proxy` to `use_istio_ingress` for the config flag. I support what @ericpromislow suggested for the refactor, but don't feel comfortable making it because I can't run CATs against `cf-for-k8s`.

This PR was initially done against a "hybrid" BOSH/Kubernetes environment where staging still happened on Diego and most of the control plane was still running on BOSH. Only `cf-k8s-networking` components and apps were running on Kubernetes.

I think we should block this on getting the test to run in `cf-for-k8s` in general. Until then this change doesn't really add any value.

jamespollard8:
after/from #429 it sounds like the plan is to refactor the changes here to use the `Config.RunningOnK8s()` helper to choose the correct expectation.

@tcdowney Does that make sense and sound good to you?

tcdowney:
@jamespollard8 

That makes sense to me. I'm no longer with the CF Networking Program, though, so someone from @cloudfoundry/cf-networking will need to take this PR over or start a new one.

Also might need to run this test again and make sure its still valid with the latest Envoys that ship in Istio.

jamespollard8:
cf k8s networking story to follow-up: https://www.pivotaltracker.com/story/show/174582818

apps/routing_transparency.go
mike1808: I understand that it's hard to see the difference, but these tests are not meant for comparison of behaviors rather than testing proxies' routing transparencies. 

The difference can be found in the test message titles. Gorouter tests assuming that everything is passed without any kind of normalization, while Istio tests assume that some unsafe characters are normalized. We'll few `By` messages to describe the assertions the tests are checking for to make this clear.

mike1808:
@jamespollard8 we fixed the test and it's ready to be merged.

PR #404, opened 2020-01-16
Remove duplicate entry from example config

gcapizzi:

The example config has a duplicated entry, which can make the behaviour of the suite very confusing if they have different values!

- [ ] YES
- [X] N/A


davewalter:
Thanks @gcapizzi!

PR #405, opened 2020-01-23
Respect Config.Protocol when testing iso segs

tlwr:

@schmie and I recently enabled the isolation segment acceptance tests in our environments, and they were failing.

They were failing because the tests made assumptions about:

- the configured protocol which should be used to test the applications
- the hostname under which the applications were accessible

We have made the protocol use the configurable protocol, which assumes that the isolation segment will be routable via gorouter, on the apps domain (which I think is a reasonable assumption).

We have changed the hostname to be the name of the application, which is consistent with the other tests.

We have used cf-acceptance-tests branch `cf12.19` against a fairly heavily modified cf-deployment `12.19`, the details of which can be found in [alphagov/paas-cf/wip-tlwr](https://github.com/alphagov/paas-cf/tree/wip-tlwr)

- [ ] YES
- [x] N/A

The isolation segments acceptance tests now respect the protocol configuration option, and use the hostname of the application, rather than a statically defined wildcard.

This should not change the runtime

@tlwr and @schmie paired on this on behalf of [GOV.UK PaaS](https://www.cloud.service.gov.uk/)

### DIFF
--- a/isolation_segments/isolation_segments.go
+++ b/isolation_segments/isolation_segments.go
@@ -124,10 +124,8 @@ var _ = IsolationSegmentsDescribe("IsolationSegments", func() {
 					"-c", "./app"),
 					Config.CfPushTimeoutDuration()).Should(Exit(0))
 
-				hostHeader := fmt.Sprintf("Host: %s.%s", appName, isoSegDomain)
-				host := fmt.Sprintf("http://wildcard-path.%s", isoSegDomain)
-
-				curlSession := helpers.CurlSkipSSL(Config.GetSkipSSLValidation(), host, "-H", hostHeader)
+				curlArgs := Config.Protocol() + appName + "." + isoSegDomain
+				curlSession := helpers.Curl(Config, curlArgs)
 				Eventually(curlSession).Should(Exit(0))
 				Expect(curlSession.Out).To(Say(binaryHi))
 			})
@@ -220,10 +218,8 @@ var _ = IsolationSegmentsDescribe("IsolationSegments", func() {
 					"-c", "./app"),
 					Config.CfPushTimeoutDuration()).Should(Exit(0))
 
-				hostHeader := fmt.Sprintf("Host: %s.%s", appName, isoSegDomain)
-				host := fmt.Sprintf("http://wildcard-path.%s", isoSegDomain)
-
-				curlSession := helpers.CurlSkipSSL(Config.GetSkipSSLValidation(), host, "-H", hostHeader)
+				curlArgs := Config.Protocol() + appName + "." + isoSegDomain
+				curlSession := helpers.Curl(Config, curlArgs)
 				Eventually(curlSession).Should(Exit(0))
 				Expect(curlSession.Out).To(Say(binaryHi))
 			})


davewalter:
Hi @tlwr & @schmie,

We are somewhat concerned by the level of modification of CF-deployment that you have tested this against. Could you please first test this change against a standard cf-d deployment to confirm that  it doesn't break the more common deployment cases?

Thanks,
Dave and @jspawar

pivotaljohn:
@tlwr & @schmie : this issue has been stale for more than 21 days. We're closing it to keep this space tidy. If you'd like to work this further, please feel free to re-open it.

tlwr:
I've been procrastinating running this against a CF development environment, my apologies.

I still intend on testing this against a CF closer to cf-deployment, when I have the time to get `cfdev` to do my bidding.

In terms of what this actually changes, I think:

- the change in this PR makes it easier to understand what is going on
- without this change, it is not clear how the tests actually work
- the change in this PR is more consistent with existing tests

jamespollard8:
> I still intend on testing this against a CF closer to cf-deployment, when I have the time to get `cfdev` to do my bidding.

Sounds good - please let us know once you've been able to do this testing.

How would you feel about simplifying this change and just changing the `host := fmt.Sprintf("http://wildcard-path` line to use `Config.Protocol`?

tlwr:
@jamespollard8 in order to answer

> How would you feel about simplifying this change and just changing the host := fmt.Sprintf("http://wildcard-path line to use Config.Protocol?

I think I need to understand a little more about how these tests work right now:

- how does `wildcard-path` get mapped to the isolation segment domain we just pushed?
- is this test actually falling through to the app defined in `apps/wildcard_routes.go`?
- is testing the `wildcard-path` rather than the app name what we actually want to test?

Personally, I think this test should look as much like the tests defined in `apps/lifecycle.go`, except test using the isolation segment domain rather than the apps domain: i.e. use the app name and the isolation segment domain, as we _explicitly_ want to test the app we just pushed.

jamespollard8:
> I think I need to understand a little more about how these tests work right now:
> 
> * how does `wildcard-path` get mapped to the isolation segment domain we just pushed? (presumably the `*` route, which isn't defined in the isolation segment test suite)
> * is this test actually falling through to the app defined in `apps/wildcard_routes.go`?
> * is testing the `wildcard-path` rather than the app name what we actually want to test?
> 
> Personally, I think this test should look as much like the tests defined in `apps/lifecycle.go`, except test using the isolation segment domain rather than the apps domain: i.e. use the app name and the isolation segment domain, as we _explicitly_ want to test the app we just pushed.

Thanks @tlwr for the response. We (Release Integration) will take a proper look when we get a chance. Please keep us in the loop if you have any relevant updates

ericpromislow:
Hi @tlwr 

We generally close PRs after 21 days. Please feel free to create a new PR for this once you're able to test it against a recent version of `cf-deployment` .  Thanks for your understanding,

- @ericpromislow  and @Birdrock ,  Cloud Foundry Release Integration Community Pair

PR #409, opened 2020-02-18
Fix flakey loggregator test

chentom88:
Signed-off-by: Melena Suliteanu <msuliteanu@pivotal.io>

Fixes a flaky loggregator test.

https://cloudfoundry.slack.com/archives/C02HCCXV5/p1581018053047900

12.31.0

- [ ] YES
- [x] N/A

Fix flaky loggregator test

No change

@pianohacker

### DIFF
--- a/apps/default_environment_variables.go
+++ b/apps/default_environment_variables.go
@@ -96,17 +96,17 @@ exit 1
 
 			pushSessionContent := pushSession.Buffer().Contents()
 
-			Expect(pushSessionContent).To(MatchRegexp("LANG=en_US\\.UTF-8"))
-			Expect(pushSessionContent).To(MatchRegexp("CF_INSTANCE_INTERNAL_IP=.*"))
-			Expect(pushSessionContent).To(MatchRegexp("CF_INSTANCE_IP=.*"))
-			Expect(pushSessionContent).To(MatchRegexp("CF_INSTANCE_PORTS=.*"))
-			Expect(pushSessionContent).To(MatchRegexp("CF_STACK=.*"))
-			Expect(pushSessionContent).To(MatchRegexp("VCAP_APPLICATION=.*"))
-			Expect(pushSessionContent).To(MatchRegexp("VCAP_SERVICES=.*"))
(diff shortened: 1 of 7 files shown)


jamespollard8:
Guidance from logging folks:
>The PR does no harm and should almost certainly help make the tests less flaky. It should be fine to pull in.

jamespollard8:
logging some of @MasslessParticle's comments from the [slack thread](https://cloudfoundry.slack.com/archives/C02HCCXV5/p1581018053047900)

> It's strange to me that we setup the stream every time:
> ```
> s := rlpClient.Stream(ctx, ebr)
> es := s()
> ```
> My bet is that the overhead of setting this up each time is giving us plenty of time to miss tons of logs
> We'll change the test to not setup and tear down new stream once per millisecond

jamespollard8:
Looks good and simple to us

jamespollard8:
The PR diff is showing up poorly but it looks like the actually change really was just the small expected change:
```
$ git diff c0a38b5b
diff --git a/apps/loggregator.go b/apps/loggregator.go
index 0c1b0c61..c9e25123 100644
--- a/apps/loggregator.go
+++ b/apps/loggregator.go
@@ -127,15 +127,16 @@ var _ = AppsDescribe("loggregator", func() {
                                },
                        }

+                       ctx, cancelFunc := context.WithTimeout(context.Background(), Config.DefaultTimeoutDuration())
+                       defer cancelFunc()
+
+                       s := rlpClient.Stream(ctx, ebr)
+
                        Eventually(func() string {
                                return helpers.CurlApp(Config, appName, fmt.Sprintf("/log/sleep/%d", hundredthOfOneSecond))
                        }).Should(ContainSubstring("Muahaha"))

                        Eventually(func() string {
-                               ctx, cancelFunc := context.WithTimeout(context.Background(), Config.DefaultTimeoutDuration())
-                               defer cancelFunc()
-
-                               s := rlpClient.Stream(ctx, ebr)
                                es := s()
                                var messages []string
                                for _, e := range es {
```

PR #415, opened 2020-05-07
Enable docker before tests

paulcwarren:
- and disable again afterwards

Co-authored-by: Paul Warren <pwarren@pivotal.io>
Co-authored-by: Lisa Burns <lcho@pivotal.io>

_All PR's to CATs should be submitted to develop and will be merged to master once they've passed acceptance._

We are paying down some tech debt.  we wanted both pats and cats to work the same way; i.e. be responsible for enabling docker, rather than having the responsibility in the test one place and in the pipeline in another.

[#172250234](https://www.pivotaltracker.com/story/show/172250234)

a smith claimed cf-deployment

- [X] YES
- [ ] N/A

n/a

negligible.  all the test does is run two extra cf commands to enable the docker feature flag

@cloudfoundry/cf-volume-services

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ include_capi_no_bridge
 * `include_service_discovery`: Flag to include test for the service discovery. These tests use `apps.internal` domain, which is the default in `cf-networking-release`. The internal domain is currently not configurable.
 * `stacks`: An array of stacks to test against. Currently only `cflinuxfs3` stack is supported. Default is `[cflinuxfs3]`.
 
-* `include_volume_services`: Flag to include the tests for volume services. The following requirements must be met to run this suite: Diego must be deployed. Docker support must be enabled. tcp-routing must be deployed. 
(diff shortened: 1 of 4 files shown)


davewalter:
Hi @paulcwarren,

Let's talk about this. I can see potential problems when running the entire set of CATs in parallel, especially when it comes to the docker suite. For this reason, we are trying to avoid having the tests change the state of the CF deployment under test due to the potential for random failures caused by the wrong sets of tests trying to run in parallel. I think the best solution might be to add a check to prevent the volume services suite from running if the `include_docker` flag is set to avoid any possible conflicts.

Regards,
Dave

paulcwarren:
@julian-hj, @lisamburns Dave and I added a quick amendment to the PR to skip volume services when docker tests are also enabled so there is no chance of our tests putting the platform in a state that would cause the docker tests to fail.

However, given there is quite a bit of setup to do we decided to run it through the pipeline instead.

PR #422, opened 2020-07-09
Migrate CATs to use v7 CF CLI

acosta11:
# DO NOT MERGE (YET)

Co-authored-by: Eric Promislow <epromislow@suse.com>

Yes

While moving CATs to use CF CLI 7, we're maintaining the changes as a PR to make it easier to 
solicit community feedback

https://www.pivotaltracker.com/epic/show/4320018

We actually ran this through the CATs ci before reverting and open a PR, which used the latest release: 13.6.0 

- [x] YES, but may warrant additional changes based on feedback.
- [ ] N/A

The v7 cli and v3 api include new functionality, for which we would like new acceptance tests.

Update CATs to reflect the use of cf cli v7 instead of v6 going forward.

N.A.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,6 @@ include_apps
 include_detect
 include_routing
 include_v3
-include_capi_no_bridge
 ```
 
 #### The full set of config parameters is explained below:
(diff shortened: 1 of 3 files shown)


acosta11:
Messed up the merge commit on the branch, so re-opened with correct diff in #423

PR #423, opened 2020-07-10
Migrate CATs to use v7 CF CLI

acosta11:
## Migrate CATs to use v7 CF CLI

Co-authored-by: Eric Promislow <epromislow@suse.com>
Co-authored-by: Dave Walter <dwalter@pivotal.io>

Yes

While moving CATs to use CF CLI 7, we're maintaining the changes as a PR to make it easier to 
solicit community feedback

### List of Changes
- Stop using `-d` flag
  - We make the concession of an app manifest in places where the domain matters
- Stop using the log cache plugin in cli and directly hook to logs
  - We now use the `Recent()` helper instead of `Tail()`
  - Using `Follow()` instead of `TailFollow()`
- Stop using incompatible buildpack index
  - Anywhere that buildpack index was 0, we update to 1
- Use a `—space` flag instead of positional arg for security groups
-`run-task` now takes a `—command` flag before the command
- Healthchecks no longer allow the use of `none` type, instead requiring an explicit type
  - Removed redundant windows test using the `none` type
- Healthcheck now takes an endpoint
  - Added the `-l` flag to specify endpoint with healthchecks
- Adding a network policy now uses a positional arg instead of the `—destination-app` flag
- Added a sleep command for test stability on functionality known to work quickly when run manually
  - Looks like an artifact of the underlying implementation of `Eventually()`

https://www.pivotaltracker.com/epic/show/4320018

We actually ran this through the CATs ci before reverting and open a PR, which used the latest release: 13.6.0 

- [x] YES, but may warrant additional changes based on feedback.
- [ ] N/A

The v7 cli and v3 api include new functionality, for which we would like new acceptance tests.

Update CATs to reflect the use of cf cli v7 instead of v6 going forward.

N.A.


windows/instance_reporting.go
```diff
@@ -48,7 +48,7 @@ var _ = WindowsDescribe("Getting instance information", func() {
 			setTotalMemoryLimit("10G")
 		})
 
-		It("fails when scaled beyond available resources", func() {
+		PIt("fails when scaled beyond available resources", func() {
```
ericpromislow, line 51: Pending test - looks like an artefact

windows/lifecycle.go
```diff
@@ -26,15 +26,15 @@ var _ = WindowsDescribe("Application Lifecycle", func() {
 		appName = random_name.CATSRandomName("APP")
 	})
 
-	It("exercises the app through its lifecycle", func() {
+	PIt("exercises the app through its lifecycle", func() {
```
ericpromislow, line 29: Pending test

windows/set_start_command.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
 )
 
-var _ = WindowsDescribe("Setting an app's start command", func() {
+var _ = FDescribe("Setting an app's start command", func() {
```
ericpromislow, line 17: Focused describe-group

windows/set_start_command.go
```diff
@@ -38,10 +38,11 @@ var _ = WindowsDescribe("Setting an app's start command", func() {
 		Expect(cf.Cf("delete", appName, "-f", "-r").Wait()).Should(Exit(0))
 	})
 
-	It("uses the given start command", func() {
+	// Couldn't find app, placing into pending for now
+	PIt("uses the given start command", func() {
```
ericpromislow, line 42: Pending test

apps/piecemeal_app_creation_flow.go
```diff
@@ -0,0 +1,75 @@
...
+
+	It("successfully creates and starts a new app with discrete commands", func() {
+		Expect(cf.Cf("create-app", appName).Wait()).To(Exit(0))
+		session := cf.Cf("app", appName, "--guid")
+		Eventually(session).Should(Exit(0))
+		appGUID := strings.TrimSpace(string(session.Buffer().Contents()))
+
+		Expect(cf.Cf("create-package", appName, "-p", assets.NewAssets().Dora).Wait()).To(Exit(0))
+
+		path := fmt.Sprintf("v3/apps/%s/packages", appGUID)
```
ericpromislow, line 43: Why not call `cf packages appName` instead?

windows/instance_reporting.go
acosta11: Unpended in subsequent update.

windows/lifecycle.go
acosta11: Unpended in subsequent update.

windows/set_start_command.go
acosta11: Unfocused in subsequent update.

apps/piecemeal_app_creation_flow.go
acosta11: Confirmed our suspicion that the use of curl was for ease of response parsing.

acosta11:
Confirmed all checks have passed aside from the CLA, for which we know Julian is a member of the team. Will get the CLA sorted out afterwards. Going to go ahead and merge now that we've waited sufficiently for community feedback.

PR #429, opened 2020-07-31
Skip tests on k8s

acosta11:

Add the ability to control which tests will run based on the configured infrastructure ("kubernetes" or "vms")

https://www.pivotaltracker.com/story/show/174026222

v13.11.0

- [x] requires an update to a CATs integration-config (if you are targeting a cf-for-k8s environment)

- [x] YES
- [ ] N/A

Adds the ability to control which tests will run based on the configured infrastructure ("kubernetes" or "vms")

Skipping tests should make CATs run faster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,8 @@ include_v3
 * `volume_service_plan_name`: The name of the plan of the service provided by the volume service broker.
 * `volume_service_create_config`: The JSON configuration that is used when volume service is created.
 
+* `infrastructure`: The name of the infrastructure for the environment that the tests will run against. Must be either "vms" or "kubernetes". Defaults to "vms".
+
 #### Buildpack Names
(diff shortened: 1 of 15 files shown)


davewalter:
> Nice - looks clean 👍 . I'm curious what the plan is for ordering / integrating these changes with the go-router vs istio changes from #396

Our plan was for them to use the `Config. RunningOnK8s()` helper to choose the correct expectation.

ericpromislow:
After applying this change we couldn't actually get any cats tests to pass when targeting a cf-for-k8s foundation.  We ran into incompatibilities like specifying buildpacks, which cf-for-k8s doesn't currently allow.

davewalter:
> After applying this change we couldn't actually get any cats tests to pass when targeting a cf-for-k8s foundation. We ran into incompatibilities like specifying buildpacks, which cf-for-k8s doesn't currently allow.

I don't think this PR was intended to fix all the issues with running CATs against a cf-for-k8s environment. This was just adding the infrastructure to allow us to mark specific tests as not being suitable to run. The tests we did flag in this PR were just a "proof of concept".

PR #432, opened 2020-08-27
174257193 cats continued

Birdrock:

Updating CATS (app suite) to work with CF on Kubernetes.

https://www.pivotaltracker.com/story/show/174257193

- [X] requires an update to a CATs integration-config

It is now possible to run a subset of CATS against a CF on Kubernetes!

@Syerram @jamespollard8

### DIFF
--- a/apps/changing_start_command.go
+++ b/apps/changing_start_command.go
@@ -35,8 +35,7 @@ var _ = AppsDescribe("Changing an app's start command", func() {
 	Context("by using the command flag", func() {
 		var expectedNullResponse string
 
-		// TODO: No binary buildpack w/o Procfile support
-		SkipOnK8s()
+		SkipOnK8s("Custom start command not supported")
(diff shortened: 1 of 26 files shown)


apps/changing_start_command.go
```diff
@@ -35,6 +35,9 @@ var _ = AppsDescribe("Changing an app's start command", func() {
 	Context("by using the command flag", func() {
 		var expectedNullResponse string
 
+		// TODO: No binary buildpack w/o Procfile support
```
acosta11, line 38: Is the idea here that we don't have support for the -c flag with a procfile regardless? And the cli previously only allowed us to use the -c flag with the binary buildpack?

davewalter: I think it is more that we simply don't have support for custom start commands at all. You have to use a Procfile.

PR #433, opened 2020-08-28
Fix flaky wildcard routing test

mdelillo:
This test fails if it uses a CATS-org that has more than one space. Explicitly targeting the space fixes it.


davewalter:
Hi @mdelillo,

Thanks for the contribution. We already have this change on the `release-candidate` branch via [this commit](https://github.com/cloudfoundry/cf-acceptance-tests/commit/5b9efd2cd8a9977d84d387b3de76c31bab6f5e23).

We're working on making CATs compatible with cf-for-k8s (see #432) and will cut a new release once that is merged.

Thanks,
Dave and @Birdrock

PR #441, opened 2020-10-08
Fix CATs to allow it to run in more configuration

jvshahid:
This PR consists of three commits that can be reviewed in isolation.  The commit message of each change explains the intention behind the change.

We intend to run CATs in our managed environment on AWS.

https://vmware.slack.com/messages/C3LUD2L04/p1602100943001300

CATs 5.11.0

- [ ] YES
- [X] N/A

I think the three commit messages are good candidates since they are descriptive and brief.  I will include them below:

```
Use https syslog drain URL instead of syslog-tls.
Use Config.Protocol() instead of hard coding the protocol.
Fix the test to reflect the original intention.
```

This change will make CATs faster since the Syslog test was simplified and doesn't wait or parse the Syslog listener logs anymore.

@Logiraptor @houlistonm

### DIFF
--- a/apps/syslog_drain.go
+++ b/apps/syslog_drain.go
@@ -60,7 +60,6 @@ var _ = AppsDescribe("Logging", func() {
 			Eventually(cf.Cf(
 				"push",
 				logWriterAppName2,
-				"--no-start",
 				"-b", Config.GetRubyBuildpackName(),
 				"-m", DEFAULT_MEMORY_LIMIT,
 				"-p", assets.NewAssets().RubySimple,
--- a/apps/syslog_drain.go
+++ b/apps/syslog_drain.go
@@ -1,7 +1,6 @@
 package apps
 
 import (
-	"regexp"
 	"time"
 
 	. "github.com/cloudfoundry/cf-acceptance-tests/cats_suite_helpers"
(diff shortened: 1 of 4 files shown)


davewalter:
Hi @jvshahid,

We are seeing a [failure in our CATs CI pipeline](https://release-integration.ci.cf-app.com/teams/main/pipelines/cats/jobs/run-cats/builds/249) with these changes. The syslog drains test (`apps/syslog_drain.go:98`) is consistently failing to validate that the logs contain the appropriate random message from the app that is supposed to be logging output. We have manually verified that this is related to the change to the commit to "Use https syslog drain URL instead of syslog-tls". Our pipeline is using a toolsmiths' pooled environment running the latest release of cf-deployment on GCP. Can you think of any reason why this should be failing?

Regards,
Dave

jvshahid:
I'm sorry about that. I'm away from a computer at the moment but I could
take a look tomorrow morning.

jvshahid:
To answer your question, it looks like the https load balancer to the
gorouter is unreachable from inside the diego containers. I'm not sure why
though. You could check the loggr-syslog-agent logs for more information.
It could be a security group issue or something with the infrastructure
firewall rules.

davewalter:
Thanks @jvshahid. We'll take another look at it on Monday.

davewalter:
Hi @jvshahid,

To get CI green, we have decided to revert 3fa457a886cadc367ffdaa0f2ceb2ec9a96e0700 as that is the change that is causing problems. If it is a change you still want to incorporate into CATs, it would be great if you could open a new PR and work with the networking folks to get this working. We're also available to help with specifics around our CI environments.

Regards,
Dave and @jamespollard8

PR #447, opened 2020-11-17
Split the syslog drain test into a separate suite

davewalter:

Yes

This change moves the `syslog drain` test to a separate suite to allow it to be skipped while still running the majority of the `apps` suite.

The original request to update this test to always use TLS came from PR #441. This was reverted via https://github.com/cloudfoundry/cf-acceptance-tests/commit/7e1089ca9c538664f3b77293e53380b954abc1dc due to CI failures. After discussions with @jvshahid, we decided to move the test to a separate suite to allow it to be skipped.

v15.0.0

- [x] YES
- [ ] N/A

Moved the syslog drain test to a new suite to allow it to be skipped on environments that do not allow internal TCP traffic.

@jamespollard8

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ include_v3
 
 ##### Optional parameters:
 `include_*` parameters are used to specify whether to skip tests based on how a deployment is configured.
+* `include_app_syslog_tcp`: Flag to include the app syslog drain over TCP test group.
 * `include_apps`: Flag to include the apps test group.
 * `include_container_networking`: Flag to include tests related to container networking.
   `include_security_groups` must also be set for tests to run. [See below](#container-networking-and-application-security-groups)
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -338,6 +339,7 @@ You can of course combine the `-v` flag with the `-nodes=N` flag.
 
 Test Group Name| Description
 --- | ---
+`app_syslog_gcp`| Tests the ability to configure an app syslog drain listener.
```
acosta11, line 342: Should this be app_syslog_ `tcp` instead of `gcp`?

acosta11:
Hi Dave,

It looks like we should just rebase this on develop and fix the typo in the README, then this will be good to merge.

Thanks,
Andrew and Eric

PR #448, opened 2020-11-23
Remove check on 'binding in progress' message for async bind/unbind as

pivotal-marcela-campo:
we do not print it if the job is already completed when we first poll.

[#175617691](https://www.pivotaltracker.com/story/show/175617691)

Remove check on "Binding in progress" message from async bind/unbind.

When v8 cli gets a 202 response for binding/unbinding, it first polls the api job for status and doesn't print any messages until the job is either in polling state or completed.

In the current broker implementation, last operation for service bindings will always return completed (never in progress) so the whole job finishes almost instantly and it's completed by the time the cli requests the last operation status.

The "binding in progress" message only gets printed if the job is in the 'polling' state when the last operation endpoint is queried. But in the current CATS implementation that never happens. Hence removing the message check.

We are proving anyway that the cli/api can handle async broker responses as we are using an async plan and the broker is responding with a 202. It is just a different path for the messaging that would require implementing something in the lines of 'max_fetch_service_instance_requests' but for service bindings if we wanted to check both paths.

- [ ] YES
- [X] N/A

_Something brief that conveys the change and is written with the component author audience in mind._

### DIFF
--- a/services/service_instance_lifecycle.go
+++ b/services/service_instance_lifecycle.go
@@ -518,7 +518,6 @@ var _ = ServicesDescribe("Service Instance Lifecycle", func() {
 				bindService := cf.Cf("bind-service", appName, instanceName).Wait()
 
 				Expect(bindService).To(Exit(0), "failed to asynchronously bind service")
-				Expect(bindService).To(Say("Binding in progress."))
 
 				By("waiting for binding to be created")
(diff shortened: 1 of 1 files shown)


jamespollard8 (approved):
Thanks a bunch - this looks great to us!

James and @Birdrock

PR #449, opened 2020-12-04
Remove the internetless test suite from CATs

ericpromislow:
[#175809268](https://www.pivotaltracker.com/story/show/175809268)

This is a simple change -- just deleting a test and all the supporting apparatus from cats, but I can't run a service-free subset of cats without errors, so made a PR


jamespollard8 (approved):
LGTM!

PR #450, opened 2020-12-10
Use vendir to vendor the logging-route-service sample app

jvshahid:

Yes

This PR changes the way we vendor `assets/logging-route-service/`.  Instead of using git submodules, the assets are now vendored using [vendir](https://github.com/vmware-tanzu/carvel-vendir).  This change makes it possible to download a self-contained CATs zip file from the Github releases page.  Previously the CATs zip file would be missing the asset mentioned before.

https://vmware.slack.com/archives/C3LUD2L04/p1606239248053300

I didn't run the tests.

- [ ] YES
- [X] N/A

Vendor `assets/logging-route-service` using [vendir](https://github.com/vmware-tanzu/carvel-vendir).

Zero.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "assets/logging-route-service"]
-	path = assets/logging-route-service
-	url = https://github.com/cloudfoundry-samples/logging-route-service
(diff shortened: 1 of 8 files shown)


jamespollard8 (approved):
Cool - this looks great!

I was thinking about a strategy for making sure this doesn't go stale but then I saw that the source repo hasn't been updated in a year so I think we're safe there 😅

PR #456, opened 2021-02-02
UAA v75 CAT flakes

bruce-ricard:
* The test was expecting the UAA authorization code to only contain
  letters and numbers, which was a incorrect expectation. For example,
  as of today, the code can contain dashes "-".

* The regex was updated so that the code can contain any character. If
  the UAA decides one day to add another CGI parameter to the redirect
  URL, the test won't break because of the "^&" in the regex.

* The '\r' and '\n' are there to end the regex match at the end of the
  line.

* For context, this is what an UAA API response looks like. See line 6
  the code we are trying to extract.

HTTP/2 302
cache-control: no-store
content-language: en-US
content-length: 0
date: Tue, 02 Feb 2021 22:10:17 GMT
location: http://example.com?code=gfCMdm9L3P                # <- code here
strict-transport-security: max-age=31536000
x-content-type-options: nosniff
x-frame-options: DENY
x-vcap-request-id: 82abfda5-bf9c-4790-726a-b45d20d8883c
x-xss-protection: 1; mode=block
via: 1.1 google
alt-svc: clear

Yes!

Fixing the CATs compatibility with UAA.

https://cloudfoundry.slack.com/archives/C03FXANBV/p1612201884012700

latest

- [ ] YES
- [x] N/A

0

@reedr3 , @peterhaochen47

### DIFF
--- a/helpers/services/sso.go
+++ b/helpers/services/sso.go
@@ -117,7 +117,7 @@ func AuthorizeScopes(cookie string, config OAuthConfig) (authCode string) {
 	Expect(curl).To(Exit(0))
 	apiResponse := string(curl.Out.Contents())
 
-	pattern := fmt.Sprintf(`%v\?code=([a-zA-Z0-9]+)`, regexp.QuoteMeta(config.RedirectUri))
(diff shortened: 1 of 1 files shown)


jamespollard8 (approved):
LGTM - thanks again!

PR #461, opened 2021-03-08
disable GO111MODULE for logging route service test app

peterellisjones:

Yes

Go buildpack [v1.9.27](https://github.com/cloudfoundry/go-buildpack/releases/tag/v1.9.27) sets `GO111MODULE=auto` by default (note the release notes state `GO111MODULES` but the actual env var is [`GO111MODULE`](https://blog.golang.org/modules2019) without the 's'). As the logging route service test app does not contain a `go.mod` file it now requires `GO111MODULE=off` to be set explicitly for staging to be successful.

Note that `GO111MODULE: off` and `GO111MODULE: "off"` are not equivalent YAML because `off` is a boolean, not string YAML value, hence the need for quotes in the app manifest.

Go buildpack v1.9.27 is not yet in a cf-deployment version, but I have tested this against a Cloud Foundry based on cf-deployment v16.5.0 running Go buildpack v1.9.27. Setting `GO111MODULE=off` explicitly should not alter functionality of the test app for any other versions of the Go buildpack as it will either be ignored, or would have automatically been set to `off` anyway.

- [ ] YES
- [x] N/A

n/a

_Update logging route service test app for Go buildpack v1.9.27 compatibility_

0

n/a

### DIFF
--- a/assets/logging-route-service/manifest.yml
+++ b/assets/logging-route-service/manifest.yml
@@ -5,3 +5,4 @@ applications:
   env:
     GOVERSION: 1.x
     GOPACKAGENAME: github.com/cloudfoundry-samples/logging-route-service
+    GO111MODULE: "off"


Birdrock:
@peterellisjones Thanks for the PR! Ultimately, we decided to fix this behavior by updating the assets to contain a go.mod file, so we are closing this PR unmerged, but we appreciate your assistance!

PR #462, opened 2021-03-08
Replaced all v2 API calls with v3 calls

FelisiaM:
[#177245959](https://www.pivotaltracker.com/story/show/177245959)

Services test relly only on v3 CAPI endpoints.

v16.5.0

Services test relly only on v3 CAPI endpoints.

### DIFF
--- a/services/service_instance_lifecycle.go
+++ b/services/service_instance_lifecycle.go
@@ -26,19 +26,10 @@ type LastOperation struct {
 	State string `json:"state"`
 }
 
-type Entity struct {
-	Name          string        `json:"name"`
-	LastOperation LastOperation `json:"last_operation"`
-}
-
-type Metadata struct {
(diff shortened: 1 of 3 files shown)


services/service_instance_lifecycle.go
```diff
@@ -288,13 +279,13 @@ var _ = ServicesDescribe("Service Instance Lifecycle", func() {
 					Expect(bindService).To(Exit(0), "failed binding app to service")
 				})
 
-				Context("when there is an existing binding", func() {
+				FContext("when there is an existing binding", func() {
```
matt-royal, line 282: I don't think you meant to leave this focus (and the one below) in

FelisiaM: Thanks for catching it. I removed it now

PR #467, opened 2021-05-06
Add a test to ensure uaa/capi can emit metrics about system health

Benjamintf1:
- used to ensure component `statsd injector release` is working

Signed-off-by: Josh Russett <jrussett@vmware.com>

We want to run CATs to ensure the health/development of specifically the statsd injector. You can delete the statsd injector from the platform and no other tests fail, thusly we want to ensure increased coverage. Statsd injector is used to emit metrics about uaa and capi into loggregator. This helps us deprecate cfarlats

latest

- [x] introduces a new test --- Are you sure everyone should be running this test? everyone has a uaa, it should be mostly equivilent to previous tests

- [x] YES
- [ ] N/A

operators like their metrics coming out of loggregator

new test to validate that uaa and capi can emit metrics about themselves

less then 10 seconds

@cloudfoundry/runtime

### DIFF
--- a/apps/loggregator.go
+++ b/apps/loggregator.go
@@ -113,7 +113,7 @@ var _ = AppsDescribe("loggregator", func() {
 	Context("reverse log proxy", func() {
 		SkipOnK8s("Not yet supported")
 
-		It("streams logs", func() {
+		It("streams app logs", func() {
 			rlpClient := loggregator.NewRLPGatewayClient(
 				getLogStreamEndpoint(),
 				loggregator.WithRLPGatewayHTTPClient(newAuthClient()),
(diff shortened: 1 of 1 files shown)


Benjamintf1:
hmm, i think we @ logging and metrics should maybe put more thought into this.

heycait:
> hmm, i think we @ logging and metrics should maybe put more thought into this.

@Benjamintf1 is this ready to go in? Or is there still more here so we shouldn't merge?

Benjamintf1:
@heycait I'm rethinking this. As much as I'd love to standardize our pipes around cats, I am unsure(starting to think it's a bad idea) to introduce more "non app dev"/"platform expirience" queries/tests into cats. But I'd be interested in your thoughts here if you wish.

heycait:
Yea that makes sense not to add more to CATs. Also to note, we are just the current custodians of CATs so this question might be better answered by folks dealing with CF directly

PR #473, opened 2021-07-15
Add CATs for HTTP/2 and gRPC apps

Gerg:
### Are you submitting this PR against the [develop branch](https://github.com/cloudfoundry/cf-acceptance-tests/tree/develop)?

Yes

### What is this change about?

Add CATs for HTTP/2 and gRPC routing for applications.

### Please provide contextual information.

Root issue: cloudfoundry/routing-release#200

This is part of a larger effort to add HTTP/2 support for app routing. 

### What version of cf-deployment have you run this cf-acceptance-test change against?

v16.17.0 on a bosh lite and v16.18.0 on a full (GCP) deployment

### Please check all that apply for this PR:
- [x] introduces a new test --- Are you sure everyone should be running this test?
- [ ] changes an existing test
- [x] requires an update to a CATs integration-config

### Did you update the README as appropriate for this change?
- [x] YES
- [ ] N/A

### If you are introducing a new acceptance test, what is your rationale for including it CATs rather than your own acceptance test suite?

HTTP/2 support is core functionality for Cloud Foundry. Its implementation spans multiple releases (at least `capi-release`, `diego-release`, and `routing-release`).

### How should this change be described in cf-acceptance-tests release notes?

Add tests for HTTP/2 and gRPC routing.

### How many more (or fewer) seconds of runtime will this change introduce to CATs?

```
+ ginkgo
Running Suite: CATS
===================
Random Seed: 1626389364
Will run 2 of 213 specs

Running CATs with CF CLI version  cf version 7.2.0+be4a5ce2b.2020-12-10

SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS
------------------------------
• [SLOW TEST:64.028 seconds]
[HTTP/2 routing]
/home/pivotal/workspace/cf-acceptance-tests/cats_suite_helpers/cats_suite_helpers.go:129
  HTTP/2 Routing
  /home/pivotal/workspace/cf-acceptance-tests/cats_suite_helpers/cats_suite_helpers.go:135
    when a destination only supports HTTP/2
    /home/pivotal/workspace/cf-acceptance-tests/http2_routing/http2_routing.go:27
      routes traffic to that destination over HTTP/2
      /home/pivotal/workspace/cf-acceptance-tests/http2_routing/http2_routing.go:28
------------------------------
[...]

PR #474, opened 2021-08-11
[#178937872] Adds TCP routing tests to apps on isolation segments

geofffranks:

Adds tests to ensure that TCP routed apps work on isolation segments (but not for routed isolation-segments)

https://www.pivotaltracker.com/story/show/178937872

https://github.com/geofffranks/cf-deployment/tree/tcp-isolation / https://github.com/cloudfoundry/cf-deployment

Only runs the new test when both tcp_routing + isolation_segment suites are included

- [X] YES
- [ ] N/A

This test is only enabled when both tcp_routing and isolation_segments are included, and seems the best place to test a CloudFoundry environment using both of these features.

Adds tests to validate that TCP routed applications function when deployed on a shared-routing isolation segment.

~300 seconds (3 tests, 2 app pushes) when both TCP routing + isolation segments are included in the test.

### DIFF
--- a/cats_suite_helpers/cats_suite_helpers.go
+++ b/cats_suite_helpers/cats_suite_helpers.go
@@ -2,6 +2,8 @@ package cats_suite_helpers
 
 import (
 	"fmt"
+	"net"
+	"regexp"
 	"strings"
 	"time"
 
@@ -53,6 +55,17 @@ func AppsDescribe(description string, callback func()) bool {
 		BeforeEach(func() {
 			if !Config.GetIncludeApps() {
 				Skip(skip_messages.SkipAppsMessage)
+			}
+		})
+		Describe(description, callback)
+	})
+}
+
+func IsolatedTCPRoutingDescribe(description string, callback func()) bool {
+	return Describe("[tcp routing]", func() {
+		BeforeEach(func() {
+			if Config.GetIncludeRoutingIsolationSegments() || !Config.GetIncludeTCPRouting() || !Config.GetIncludeIsolationSegments() {
+				Skip(skip_messages.SkipIsolatedTCPRoutingMessage)
 			}
 		})
 		Describe(description, callback)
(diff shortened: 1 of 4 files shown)


Birdrock:
lgtm - requested review from @acosta11 for additional eyes.

acosta11 (approved):
lgtm as well

a-b:
This issue blocks the CF CLI v8 release. If we don't hear from you soon we will have to temporarily revert this PR to let you fix it without pressure.

davewalter:
Thanks for reporting your issue @a-b. I can't see the output from either of the CI builds you linked, but this change actually [broke our CATs as well](https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-deployment/jobs/upgrade-cats/builds/3204). Could you please compare the failures in our CI to yours and let us know if they are the same?

@geofffranks, I am going to revert the change from this PR as I need to merge #473 and fix an issue with another test. I would be happy to revisit this once the other changes are in and validated.

geofffranks:
@a-b I can't see the output either, but looking at https://github.com/cloudfoundry/cli-ci/blob/master/ci/cli-v8/tasks/cats-config.yml, my guess would be that you might not have DNS that points `tcp.<apps-domain>` to loadbalancers fronting the TCP router on the environment running CATs.

@davewalter I'm not sure how this would have broken https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-deployment/jobs/upgrade-cats/builds/3204. The errors there looks like issues with the CLI being unable to ssh into containers, or set environment variables across multiple test suites. This change only affects the isolation-segment + tcp routing suites.

davewalter:
> @davewalter I'm not sure how this would have broken https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-deployment/jobs/upgrade-cats/builds/3204. The errors there looks like issues with the CLI being unable to ssh into containers, or set environment variables across multiple test suites. This change only affects the isolation-segment + tcp routing suites.

I'm not sure either. The thing that really confused me was the nature of the failures. For example:

> [2021-08-23 16:52:47.96 (UTC)]> cf push CATS-9-APP-9ffb25b2f4ff5e9d -m 256M -p assets/php 
Pushing app CATS-9-APP-9ffb25b2f4ff5e9d to org CATS-9-ORG-8eef42608244b4da / space CATS-9-SPACE-f17a0406d9e402e9 as CATS-9-USER-a0aa39aaa4518a53...
FAILED
Space not found

I ran the job 5 times and saw the same (or similar) failures each time. As soon as I pinned the version of CATs back to the previous commit, they passed first time.

I will be happy to dig into it further once #473 is merged and validated.

geofffranks:
@a-b paired with us today to show us the errors in the cf-cli pipeline. They're all over the board as well. I'm wondering if there's something with cleanup interplay or bad things with scope related to the private methods i moved to be helpers that broke things.

Either way there are environmental concerns to getting the tests running, and rather than try to have everyone else fix their environments and set everything up properly, I'm wondering if we should just abandon this test in cats and move it to something like rats, adding isolation segment support there.

PR #478, opened 2021-09-10
Re-add isolated tcprouting tests with a standalone option

geofffranks:

Yes

Re-implements the #474 to play nicer with other environments. Adds a retargeting step in isolation segment after each calls to ensure other tests running get targeted to the correct space. Enables testing of TCP routing to isolated diego cells.

Re-implements #474 

latest

New property (optional, default off) to run `include_tcp_isolation_segments`. If enabled, requires `include_isolation_segments` and `isolation_segment_name`

- [ ] YES
- [x] N/A

CATs validates the proper configuration of isolated diego-cells in conjunction with a shared tcp router. It's best to test this in a scope responsible for the shipped configuration of both components.

Enables optional testing of TCP routing on isolation-segments via the `include_tcp_isolation_segments` config property.

300s when enabled

- [ x ] **Slightly Less than Urgent**


davewalter:
Thanks for the updated PR @geofffranks! Looks good.

PR #479, opened 2021-09-13
Fix incorrectly gitignore'd vendor files for HTTP/2

Gerg:

- The gitignore files for the http2 and grpc assets were ignoring the
compiled binary names, which happened to collide with the names of
vendored golang modules.
- Because the gitignore files are applied recursively, git was ignoring
these modules, which meant the apps did not work unless you ran `go mod
vendor` before pushing
- We updated the gitignore files to only apply to the compiled binary
files in the sample apps' root directories

Root issue: cloudfoundy/routing-release#200
Previous PR: https://github.com/cloudfoundry/cf-acceptance-tests/pull/473

v16.23.0

- [ ] YES
- [x] N/A

N/A

N/A

N/A

- [] **Slightly Less than Urgent**

@MerricdeLauney


davewalter:
Thanks @Gerg!

PR #484, opened 2021-11-02
Add tests for user-provided services

andy-paine:
Add tests for user-provided service operations

As part of fixing https://github.com/cloudfoundry/cloud_controller_ng/issues/2543 it is useful to have some CATs that exercise the user-provided service behaviour. In particular the test that service bindings get updated when the service gets updated will currently fail on `cf8` due to the CC v3 endpoint not having the correct behaviour.

Given these are dedicated user-facing commands on the CLI (`cf cups` and `cf uups`), I figured it made sense to have some CATs to cover their behaviour

Latest (v16.25.0)

- [x] introduces a new test --- Are you sure everyone should be running this test? Yes, they are standard behaviour and they are fast
- [x] requires an update to a CATs integration-config - this felt like a new set of tests but I could understand if you would prefer them to fall under the existing service tests flag

- [x] YES
- [ ] N/A

Add tests for user-provided service operations

~60s with 4 nodes (slowest test is 60 seconds, fastest are <2 seconds)

Working with team at SAP (can contact @philippthun, @FloThinksPi or @stephanme for followup)

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ include_v3
 * `include_sso`: Flag to include the services tests that integrate with Single Sign On. `include_services` must also be set for tests to run.
 * `include_tasks`: Flag to include the v3 task tests. `include_v3` must also be set for tests to run. The CC API task_creation feature flag must be enabled for these tests to pass.
 * `include_tcp_routing`: Flag to include the TCP Routing tests. These tests are equivalent to the [TCP Routing tests](https://github.com/cloudfoundry/routing-acceptance-tests/blob/master/tcp_routing/tcp_routing_test.go) from the Routing Acceptance Tests.
+* `include_user_provided_services`: Flag to include test for user-provided services.
 * `include_v3`: Flag to include tests for the v3 API.
 * `include_zipkin`: Flag to include tests for Zipkin tracing. `include_routing` must also be set for tests to run. CF must be deployed with `router.tracing.enable_zipkin` set for tests to pass.
 * `use_http`: Set to true if you would like CF Acceptance Tests to use HTTP when making api and application requests. (default is HTTPS)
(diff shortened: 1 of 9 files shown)


JenGoldstrich (commented):
Hey Andy!

Thanks for submitting this PR, @ctlong and I just reviewed it, and also ran it against a 16.25 cf-deployment, all of the tests pass but one relating to the CAPI bug you mention (using uups to change the credentials parameter) Carson and I manually reproduced this bug on newest cf-deployment, I assume we just need to wait for a capi-release with this PR https://github.com/cloudfoundry/cloud_controller_ng/pull/2558?

Three other small things outside of the things we marked in review that we would ask you to change

- Adding `include_user_provided_services` to the [example cats config](https://github.com/cloudfoundry/cf-acceptance-tests/blob/develop/example-cats-config.json)
- Adding a row to the `Explanation of Test Groups` section of the readme for user provide services
- Adding unit tests to the [CATS config unit test](https://github.com/cloudfoundry/cf-acceptance-tests/blob/develop/helpers/config/config_test.go) to validate the default is set correctly, nil error handling, and setting the value.

Fundamentally makes sense to us though, and the breaking test seems like it should pass when the relevant bug is fixed on CAPI, we should be able to re-test this on an appropriate CF-Deployment and merge after that.

Thanks
Jenna and Carson

andy-paine:
Thanks @JenGoldstrich - that's really useful feedback. I think it is best to wait for a CAPI release to get these merged - in theory the cf-deployment pipelines are still using `cf7` which uses the v2 services endpoints so these tests actually pass (I'm assuming you run `cf8` locally which is where you saw the failures). These tests should pass in all scenarios though rather than just the cf-d pipelines so let's hold off until all CLI versions work 👍

ctlong:
Thanks @andy-paine, looks great!

PR #496, opened 2022-02-01
Drop -d flag no longer supported after cf6

will-gant:

Yes

Updates the setup_new_broker.rb script used to bootstrap the test service broker.

The script used the now-defunct '-d' flag for cf push, which was deprecated in version 7 of the cf CLI. As it is no longer possible to specify an app's domain via commandline flags in a 'cf push', the script has been updated to retrieve the route of the test broker after it has been deployed. It then uses this to register the broker.

n/a

- [x] YES
- [ ] N/A

n/a

n/a

n/a

@SAP/cf-controlplane

### DIFF
--- a/assets/service_broker/setup_new_broker.rb
+++ b/assets/service_broker/setup_new_broker.rb
@@ -7,19 +7,6 @@ require 'securerandom'
 broker_name = ARGV[0]
 broker_name ||= 'async-broker'
 
-env = ARGV[1]
-env ||= 'bosh-lite'
-
-env_to_domain_mapping = {
-  'bosh-lite' => 'bosh-lite.com',
-  'a1' => 'a1-app.cf-app.com',
-  'tabasco' => 'tabasco-app.cf-app.com'
-}
-
-domain = env_to_domain_mapping[env] || env
-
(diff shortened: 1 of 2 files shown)


assets/service_broker/setup_new_broker.rb
```diff
@@ -104,9 +91,17 @@ def enable_service_access
...
 
 uniquify_config
-push_broker(broker_name, domain)
 
-url = "http://#{broker_name}.#{domain}"
+push_broker(broker_name)
+
+app_guid, routes_object, url = ""
+IO.popen("cf app #{broker_name} --guid") do |cmd|
+  cmd.each { |line| app_guid = "#{line.chop}" }
```
ctlong, line 99: `cmd.each` reads a bit strangely here since the expected output is a single app guid. Also, I believe that `chomp` is generally recommended over `chop` when the goal is to remove any record separators, which I assume is the case here.

ctlong: Would you be opposed to something more like this:

```ruby
app_guid = cmd.read.chomp
```

ctlong (changes requested):

ctlong (approved):

PR #527, opened 2022-04-21
Update: firehose dependencies

ctlong:

Yes

The github.com/cloudfoundry/noaa and github.com/cloudfoundry/noaa/events packages used in CATs are from 2015, before any versions were even released. The latest github.com/cloudfoundry/noaa module is v2.1.0, and we should be using that.

Additionally:
* Most of the functions we need moved from the noaa package to the consumer package
* The events package was moved to the github.com/cloudfoundry/sonde-go module

* https://github.com/cloudfoundry/noaa
* https://github.com/cloudfoundry/cf-acceptance-tests/blob/b606d93a18f18d3140df5a12cecad45a311cee33/go.mod#L10

v19.0.0.

Haven't tested on any windows cells though.

- [x] dependency update

- [ ] YES
- [x] N/A

N/A

None.

None.


davewalter (approved):
Looks good.

PR #501, opened 2022-03-01
Add a dynamic ASG test to the security_groups test suite

reneighbor:
### Are you submitting this PR against the [develop branch](https://github.com/cloudfoundry/cf-acceptance-tests/tree/develop)?

Yes

### What is this change about?

Acceptance tests for Dynamic ASGs, to be enabled by default in cf-deployment.

### Please provide contextual information.

https://www.pivotaltracker.com/story/show/181363273

### What version of cf-deployment have you run this cf-acceptance-test change against?

v18.0.0 with silk-release update to v 3.0 and policy-server-asg-syncer enabled with this ops file: https://www.pivotaltracker.com/story/show/180837394/comments/229193630

### Please check all that apply for this PR:
- [x] introduces a new test --- Are you sure everyone should be running this test?
- [ ] changes an existing test
- [ ] requires an update to a CATs integration-config

### Did you update the README as appropriate for this change?
- [ ] YES
- [x] N/A

### If you are introducing a new acceptance test, what is your rationale for including it CATs rather than your own acceptance test suite?

xCATs should validate common operator workflows._
_CATs is not a regression test suite._
_CATs is run by every component team to validate their releases before promotion._

### How should this change be described in cf-acceptance-tests release notes?

Add a test for dynamic (no restart) security groups.

### How many more (or fewer) seconds of runtime will this change introduce to CATs?
3 mins

### What is the level of urgency for publishing this change?

- [ ] **Urgent** - unblocks current or future work
- [x] **Slightly Less than Urgent**

PR #546, opened 2022-05-13
Update proxy app to skip TLS verification

davewalter:

Yes

The Dynamic ASG test pushes an app that attempts to communicate with Cloud Controller via its internal IP address, but it doesn't trust the certificate that CC presents. This change will tell the app to not attempt to validate the certificate when making the request.

See #501 for more details.

CF v19+

- [ ] YES
- [x] N/A

N/A

N/A

@ctlong

### DIFF
--- a/assets/proxy/main.go
+++ b/assets/proxy/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -25,7 +26,7 @@ func main() {
 	mux.HandleFunc("/https_proxy/", httpsProxyHandler)
 	mux.HandleFunc("/", infoHandler(systemPort))
 
-	http.ListenAndServe(fmt.Sprintf("0.0.0.0:%d", systemPort), mux)
+	_ = http.ListenAndServe(fmt.Sprintf("0.0.0.0:%d", systemPort), mux)
 }
 
 func infoHandler(port int) http.HandlerFunc {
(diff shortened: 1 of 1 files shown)


ctlong (commented):
@davewalter could this ever work without skipping cert validation? Should it be wired up to the [`skip_ssl_validation` config option](https://github.com/cloudfoundry/cf-acceptance-tests/blob/8f5ccaf914f94c9aac9113317b4d69060e19c934/example-cats-config.json#L11) in CATs?

davewalter:
> could this ever work without skipping cert validation? 

The app is trying to communicate with cloud controller internally, which presents a cert generated by BOSH using the [`cc_tls` manifest variable](https://github.com/cloudfoundry/cf-deployment/blob/6e06b9d09aab84101ba9f5ca5aa4e8b6344cc5c7/cf-deployment.yml#L2350-L2360). For this to work, I think we would need to add `((cc_tls.ca))` to the list of trusted certs for the [rootfs](https://github.com/cloudfoundry/cf-deployment/blob/6e06b9d09aab84101ba9f5ca5aa4e8b6344cc5c7/cf-deployment.yml#L1591-L1594) and [rep](https://github.com/cloudfoundry/cf-deployment/blob/6e06b9d09aab84101ba9f5ca5aa4e8b6344cc5c7/cf-deployment.yml#L1634-L1637) jobs, but I'm not sure whether that is a good idea or not, especially since this is just to get a test to pass.

We could do this with an ops-file, but that also doesn't seem right. The alternative is rewriting the test to not attempt to communicate with an internal platform API.

ctlong (approved):
Sounds good

PR #571, opened 2022-06-27
Increase HTTP/2 Routing GRPC context timeout

markstokan:
We found that the 1 second timeout for establishing a GRPC connection was too aggressive for our Nimbus/vSphere test environments.  Increasing this timeout to 5 seconds allowed our tests to pass reliably.

We debugged this failure in a Pivotal Tracker story [here](https://www.pivotaltracker.com/story/show/182068985).

This was failing against TAS 2.11 and above.

- [ ] YES
- [x] N/A

We are running off a fork with this change but would like to return to consuming upstream cloudfoundry/cf-acceptance-tests.

@crhntr

### DIFF
--- a/http2_routing/http2_routing.go
+++ b/http2_routing/http2_routing.go
@@ -81,7 +81,7 @@ var _ = HTTP2RoutingDescribe("HTTP/2 Routing", func() {
 			tlsConfig := tls.Config{InsecureSkipVerify: true}
 			creds := credentials.NewTLS(&tlsConfig)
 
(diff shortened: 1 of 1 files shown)


davewalter (approved):
@markstokan @crhntr Thanks for the contribution. Given that this is isolated to a single test, I don't think it's a problem that we are increasing the default value for this timeout for everyone running the test. The alternative of adding a new configuration option to allow users to specify their own desired timeout seems worse to me.

PR #572, opened 2022-06-30
Upgrade go-loggregator to v9

ctlong:

Yes

Upgrade go-loggregator dependency to v9. We use this library in the apps test suite to poll Traffic Controller and RLP directly.

This is not a breaking change for CATs because it doesn't use the functionality in go-loggregator that changed in that way. However, CATs should upgrade to v9 to take advantage of the new RPC files.

None yet

- [ ] YES
- [x] N/A

N/A

Probably no change.

N/A


ctlong:
Just need to test this change quickly before I mark it as ready for review.

davewalter (approved):

PR #593, opened 2022-08-30
Test new /v3/service_instances/:guid/permissions

will-gant:

Yes

Broadly, two changes. Both are related to the sso (services) test suite, which tests the [single sign-on lifecycle for service dashboards](https://docs.cloudfoundry.org/services/dashboard-sso.html).
1. Updates the helper `QueryServiceInstancePermissionEndpoint` to hit `/v3/service_instances/:guid/permissions` instead of its v2 equivalent. We recently found that the suite would fail with v2 of the CF API disabled, and were surprised to find there was no v3 equivalent of `/v2/service_instances/:guid/permissions`, which we've now implemented (see cloudfoundry/cloud_controller_ng#2946)
2. Updates the way the tests get the CF API's info endpoint, which is needed as part of the SSO OAuth flow. The CATS service broker will now grab the CF API info endpoint from the [X-Api-Info-Location header](https://docs.cloudfoundry.org/services/supporting-multiple-cf-instances.html#x-api-info-location), which is included in all calls brokers receive from the Cloud Controller. The CATS broker now stores its value in memory, and then exposes it to the tests via a new endpoint. X-Api-Info-Location [is always](https://github.com/cloudfoundry/cloud_controller_ng/blob/c8c63950125c698ca1f93fe2c677ef0274e9a953/lib/services/service_brokers/v2/http_client.rb#L102) a url for `/v2/info`. Previously the broker ignored this header, and the test suite  was just hardcoded to hit `/info`:
https://github.com/cloudfoundry/cf-acceptance-tests/blob/145a74556952c17cbc898bfb7147dd3fc2b198a1/helpers/services/sso.go#L44
That worked only because the response objects for `/v2/info` and `/info` (the ['legacy' info endpoint](https://github.com/cloudfoundry/cloud_controller_ng/blob/c8c63950125c698ca1f93fe2c677ef0274e9a953/lib/cloud_controller/legacy_api/legacy_info.rb)) both have an `authorization_endpoint` as a top-level field

As an aside, note that the tests will now succeed with v2 disabled, as `/v2/info` is in fact [the sole v2 endpoint that continues to be served](https://github.com/cloudfoundry/capi-release/blob/35623fa6d49179deff7fd077b5cc7981fe4a9311/jobs/cloud_controller_ng/templates/nginx_external_endpoints.conf.erb#L33) even when the v2 API is disabled in the capi-release.

* cloudfoundry/cloud_controller_ng#2946 - needs to be merged before the current PR (update: this is now done)

v21.2.0, patching the CAPI release for [capi 1.135.0](https://github.com/cloudfoundry/capi-release/releases/tag/1.135.0) with the cloud_controller_ng submodule swapped out for a [branch implementing the v3 API](https://github.com/sap-contributions/cloud_controller_ng/tree/service_instance_permissions).

- [ ] YES
- [x] N/A

n/a

n/a

### DIFF
--- a/services/sso_lifecycle.go
+++ b/services/sso_lifecycle.go
@@ -25,6 +25,7 @@ var _ = ServicesDescribe("SSO Lifecycle", func() {
 		if !Config.GetIncludeSSO() {
 			Skip(skip_messages.SkipSSOMessage)
 		}
+		apiEndpoint = Config.Protocol() + Config.GetApiEndpoint()
 		broker = NewServiceBroker(
 			random_name.CATSRandomName("BRKR"),
 			assets.NewAssets().ServiceBroker,
@@ -40,10 +41,10 @@ var _ = ServicesDescribe("SSO Lifecycle", func() {
 		oauthConfig.RedirectUri = redirectUri
 		oauthConfig.RequestedScopes = `openid,cloud_controller_service_permissions.read`
 
-		apiEndpoint = Config.Protocol() + Config.GetApiEndpoint()
-		SetOauthEndpoints(apiEndpoint, &oauthConfig, Config)
-
 		broker.Create()
+
+		apiInfoEndpoint := Config.Protocol() + broker.GetApiInfoUrl() // the CF API must have already called the broker for the first time with its HTTP_X_API_INFO_LOCATION header
+		SetOauthEndpoints(apiInfoEndpoint, &oauthConfig, Config)
 	})
 
 	AfterEach(func() {
@@ -75,10 +76,11 @@ var _ = ServicesDescribe("SSO Lifecycle", func() {
 			accessToken := GetAccessToken(authCode, oauthConfig)
 
 			// use the access token to perform an operation on the user's behalf
-			canManage, httpCode := QueryServiceInstancePermissionEndpoint(apiEndpoint, accessToken, serviceInstanceGuid)
+			canRead, canManage, httpCode := QueryServiceInstancePermissionEndpoint(apiEndpoint, accessToken, serviceInstanceGuid)
 
 			Expect(httpCode).To(Equal(`200`), `The provided access token was not valid.`)
 			Expect(canManage).To(Equal(`true`))
+			Expect(canRead).To(Equal(`true`))
 		})
 	})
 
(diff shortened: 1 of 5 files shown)


will-gant:
Hi @davewalter @ctlong - would one of you mind casting your eye over this? Thanks!

services/sso_lifecycle.go
```diff
@@ -25,6 +25,7 @@ var _ = ServicesDescribe("SSO Lifecycle", func() {
 		if !Config.GetIncludeSSO() {
 			Skip(skip_messages.SkipSSOMessage)
 		}
+		apiEndpoint = Config.Protocol() + Config.GetApiEndpoint()
```
ctlong, line 28: This line seems to be unused. Can it be removed?

ctlong (changes requested):
Hey @will-gant, generally looks good! Thanks for submitting this PR.

I noticed one potentially unnecessary addition and called it out. Could you please have a look and make the change, if it makes sense.

services/sso_lifecycle.go
will-gant: Ah, so `apiEndpoint` is in fact consumed in the tests below - on [L79](https://github.com/cloudfoundry/cf-acceptance-tests/blob/ee8e82d7199fa2726cefacc2946cc0726a480533/services/sso_lifecycle.go#L79) and [L116](https://github.com/cloudfoundry/cf-acceptance-tests/blob/ee8e82d7199fa2726cefacc2946cc0726a480533/services/sso_lifecycle.go#L116).

`apiEndpoint` was previously assigned on [L43](https://github.com/cloudfoundry/cf-acceptance-tests/blob/145a74556952c17cbc898bfb7147dd3fc2b198a1/services/sso_lifecycle.go#L43-L44), and was there because its value was consumed by immediately after that on L44 by `SetOauthEndpoints`. That's no longer the case.

`SetOauthEndpoints` now takes an `apiInfoEndpoint` instead (the domain retrieved from the new [GetApiInfoUrl](https://github.com/sap-contributions/cf-acceptance-tests/blob/ee8e82d7199fa2726cefacc2946cc0726a480533/helpers/services/broker.go#L127-L137)). I guess I felt it made sense to assign `apiEndpoint` elsewhere to avoid giving the impression there's still some connection between the two. 

Since `apiEndpoint` is just a concatenation of a couple of values from config it could, of course, be assigned anywhere in `BeforeEach`. Happy to move it if you still feel it ought to be somewhere else, but hopefully this explains it.

ctlong: Gotcha, thanks for the clarification!

ctlong (approved):
LGTM

PR #601, opened 2022-09-15
Run dotnet core buildpack test with published app versions for different stacks

mariash:

Yes

dotnet core buildpack will reach NuGet registry for unpublished apps to get dotnet core library. That makes this test fail in internetless environment. Switch to using published app (e.g. vendor everything). Published apps need to be compiled for the runtime they will be running on, so there are 2 published versions included for cflinuxfs3 and cflinuxfs4. 

Makefile is provided to help rebuild those versions in case of code changes.

Slack thread https://vmware.slack.com/archives/CUR2H2KFE/p1663189449408889

e8130a2c616c4c27e215030e571dc441409f9f5b

- [ ] YES
- [X] N/A

It actually makes it faster because it skips compilation and fetching dependencies during staging.


detect/buildpacks.go
```diff
@@ -116,13 +117,19 @@ var _ = DetectDescribe("Buildpacks", func() {
...
-				return helpers.CurlAppRoot(Config, appName)
-			}).Should(ContainSubstring("Hello from dotnet-core"))
-		})
+		for _, stack := range Config.GetStacks() {
+			It(fmt.Sprintf("[%s] makes the app reachable via its bound route", stack), func() {
+				assetPath, ok := assets.NewAssets().DotnetCore[stack]
+				if !ok {
+					Fail(fmt.Sprintf("dotnet-core app is missing asset for %s stack", stack))
+				}
+				Expect(cf.Cf("push", appName, "-m", DEFAULT_MEMORY_LIMIT, "-p", assetPath).Wait(Config.DetectTimeoutDuration())).To(Exit(0))
```
ctlong, line 126: I think it makes sense to add a `-s` flag to specify the specific stack to use, right?

detect/buildpacks.go
```diff
@@ -116,13 +117,19 @@ var _ = DetectDescribe("Buildpacks", func() {
...
 	Describe("dotnet-core", func() {
-		It("makes the app reachable via its bound route", func() {
-			Expect(cf.Cf("push", appName, "-m", DEFAULT_MEMORY_LIMIT, "-p", assets.NewAssets().DotnetCore).Wait(Config.DetectTimeoutDuration())).To(Exit(0))
-
-			Eventually(func() string {
-				return helpers.CurlAppRoot(Config, appName)
-			}).Should(ContainSubstring("Hello from dotnet-core"))
-		})
+		for _, stack := range Config.GetStacks() {
+			It(fmt.Sprintf("[%s] makes the app reachable via its bound route", stack), func() {
```
ctlong, line 121: `stack` needs to be assigned to a local copy inside of the for loop or else "the It closure will capture the mutating loop variable and all the specs will run against the last element in the loop." (re: [Dynamically Generating Specs](https://onsi.github.io/ginkgo/#dynamically-generating-specs))

ctlong, line 121: Can you please either put `stack` in a Context title or rephrase the It title to something more like `makes the app reachable via its bound route in <stack>`.

detect/buildpacks.go
```diff
@@ -116,13 +117,19 @@ var _ = DetectDescribe("Buildpacks", func() {
...
+				assetPath, ok := assets.NewAssets().DotnetCore[stack]
+				if !ok {
```
ctlong, line 123: Please change this to an expectation, i.e. `Expect(ok).To(BeTrue())`. You can still add your current failure message to that [as an annotation](https://onsi.github.io/gomega/#annotating-assertions).

ctlong (changes requested):
In general, I'm a little wary of the dynamic generation approach you've taken, as it's different to the static approach generally taken in CATs, but if it works I'm open to it :).

The change to the dotnet-core app looks good to me. I just have a few comments for the detect suite changes that need to be addressed.

mariash:
I think in general cats should provide a way to test against different stacks. The configuration for CATs allows to set a list of stacks to test against, but in reality we only test against one. But this would be a bigger change and would affect many tests. As of now I only changed the dotnet test because we have different pre-compiled assets for dotnet app based on a stack.

mariash:
@ctlong it looks like I can no longer push to the branch, I opened a new PR from my fork. https://github.com/cloudfoundry/cf-acceptance-tests/pull/610

PR #607, opened 2022-09-28
Update new broker GetApiInfoUrl to support SkipSSLValidation config flag

jpalermo:

Update to #593. Without these changes we see these errors on environments that have self signed certs:
```
Summarizing 3 Failures:
  [PANICKED!] [services] SSO Lifecycle [BeforeEach] When a service broker is updated with a new dashboard client can perform an operation on a user's behalf using sso
  /usr/local/go/src/runtime/panic.go:220
  [PANICKED!] [services] SSO Lifecycle [BeforeEach] When a service broker is deleted can no longer perform an operation on a user's behalf using sso
  /usr/local/go/src/runtime/panic.go:220
  [PANICKED!] [services] SSO Lifecycle [BeforeEach] When a service broker is created with a dashboard client can perform an operation on a user's behalf using sso
  /usr/local/go/src/runtime/panic.go:220
  ```

- [ ] YES
- [X] N/A

### DIFF
--- a/helpers/services/broker.go
+++ b/helpers/services/broker.go
@@ -1,6 +1,7 @@
 package services
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


davewalter (approved):

will-gant:
My bad! Apologies. Thanks for the fix.

PR #610, opened 2022-09-30
Run dotnet core buildpack test with published app versions for different stacks

mariash:

Yes

dotnet core buildpack will reach NuGet registry for unpublished apps to get dotnet core library. That makes this test fail in internetless environment. Switch to using published app (e.g. vendor everything). Published apps need to be compiled for the runtime they will be running on, so there are 2 published versions included for cflinuxfs3 and cflinuxfs4. 

Makefile is provided to help rebuild those versions in case of code changes.

Slack thread https://vmware.slack.com/archives/CUR2H2KFE/p1663189449408889

e8130a2c616c4c27e215030e571dc441409f9f5b

- [ ] YES
- [X] N/A

It actually makes it faster because it skips compilation and fetching dependencies during staging.


detect/buildpacks.go
```diff
@@ -116,13 +117,19 @@ var _ = DetectDescribe("Buildpacks", func() {
...
 
 	Describe("dotnet-core", func() {
-		It("makes the app reachable via its bound route", func() {
-			Expect(cf.Cf("push", appName, "-m", DEFAULT_MEMORY_LIMIT, "-p", assets.NewAssets().DotnetCore).Wait(Config.DetectTimeoutDuration())).To(Exit(0))
-
-			Eventually(func() string {
-				return helpers.CurlAppRoot(Config, appName)
-			}).Should(ContainSubstring("Hello from dotnet-core"))
-		})
+		for _, stack := range Config.GetStacks() {
```
ctlong, line 120: `stack` needs to be assigned to a local copy inside of the for loop or else "the It closure will capture the mutating loop variable and all the specs will run against the last element in the loop." (re: [Dynamically Generating Specs](https://onsi.github.io/ginkgo/#dynamically-generating-specs))

ctlong (changes requested):

ctlong (approved):

jochenehret:
We've just noticed that this PR introduces a large amount of resource files to the `assets/dotnet-core/cflinuxfs3` test. This can cause problems when pushing the application. In one of our test environments, we run into a HTTP 504 gateway timeout for the `POST /v3/resource_matches` request. I had to increase the AWS load balancer idle timeout to make the test pass. Here's a link to the failed test: https://app-runtime-deployments.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/experimental-cats/builds/41

Are you aware of this problem? Can you reduce the application's resources?

PR #612, opened 2022-10-03
Print AppReport after dynamic ASG test for better debugging

dsabeti:

This change adds an app report to the end of the dynamic ASG tests. If something goes wrong with the app, it's not currently possible to know why.

v21.8.0

- [ ] YES
- [x] N/A

N/A

This change adds a single call to `cf logs --recent`, so at most a few seconds.

### DIFF
--- a/credhub/service_bindings.go
+++ b/credhub/service_bindings.go
@@ -1,6 +1,7 @@
 package credhub
 
 import (
+	"encoding/json"
 	"io/ioutil"
 	"os"
 	"path"
@@ -14,8 +15,6 @@ import (
 	. "github.com/onsi/gomega/gbytes"
 	. "github.com/onsi/gomega/gexec"
 
-	"encoding/json"
-
 	"strings"
 
 	archive_helpers "code.cloudfoundry.org/archiver/extractor/test_helper"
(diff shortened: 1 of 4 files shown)


ctlong:
@dsabeti the `CredhubDescribe` function includes a similar `BeforeEach` to `AssistedCredhubDescribe` and `NonAssistedCredhubDescribe` that should skip any setup if neither "assisted" nor "non-assisted" are set.

Also, `credhub/service_keys.go` uses a similar pattern to `credhub/service_bindings.go`, so if we one is changed then I think the other should be changed as well.

If you think https://github.com/cloudfoundry/cf-acceptance-tests/pull/612/commits/d1e1265ab33a4c7f63f7497907bff813cf9de053 is worth it, could you please also apply similar changes to `credhub/service_keys.go`. Alternatively, if you remove https://github.com/cloudfoundry/cf-acceptance-tests/pull/612/commits/d1e1265ab33a4c7f63f7497907bff813cf9de053 from this PR, then I'm happy to merge it.

ctlong:
Closing this PR due to inactivity. Please feel free to reopen it if you're still interested 😄

PR #613, opened 2022-10-03
Randomize names in rename test to avoid name collisions during parallel test runs or runs on environments with littered service instances

dsabeti:

Generates a random name for a test artifact instead of using a hard-coded one.

Sometimes when tests are canceled (e.g. by a developer using Ctrl-C), artifacts get littered. The left behind artifacts, in conjunction with not using unique names, made tests fail.

v21.8.0

- [ ] YES
- [x] N/A

N/A


ctlong (approved):
LGTM

PR #614, opened 2022-10-03
Make a more targetted assertion in the user_provided_services lifecycle test

dsabeti:

In some scenarios, CAPI returns an empty 'syslog_drain_url' field in addition to the credentials. The MatchJSON matcher is fairly strict -- if there are additional fields to what is specified, it will fail. This change makes the targeted assertion that the test actually cares about -- that the username is properly stored in the credentials.

N/A

v21.8.0

- [ ] YES
- [x] N/A

N/A

None


ctlong (approved):

ctlong:
Thanks @dsabeti ! LGTM.

PR #615, opened 2022-10-03
Reorganize credhub tests so that test setup is not run when tests are skipped

dsabeti:

Because 'AssistedCredhubDescribe' and 'NonAssistedCredhubDescribe' were nested under the top-level 'CredhubDescribe', the BeforeEach of the top-level describe block would be run even when Assisted or NonAssisted tests were skipped. This removes the top-level CredhubDescribe and uses helper functions for the set up functionality shared by the two types of tests.

v21.8.0

- [ ] YES
- [x] N/A

N/A

This test saves as much as 3 minutes and 20 seconds from the test suite. The unnecessary test setup takes about 100 seconds, so in cases where all credhub tests are skipped, this saves about 200 seconds or 3 minutes and 20 seconds.

### DIFF
--- a/credhub/service_bindings.go
+++ b/credhub/service_bindings.go
@@ -1,6 +1,7 @@
 package credhub
 
 import (
+	"encoding/json"
 	"io/ioutil"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


ctlong:
@dsabeti it seems like #612 includes the changes in this PR. Can I close this PR?

ctlong:
Closing this PR due to inactivity. Please feel free to reopen it if you're still interested 😄

PR #620, opened 2022-10-12
Fix: log rate limiting lifecycle tests

mkocher:

Yes.

* Apps that aren't logging anything don't show a log rate greater than zero.
  * Added a call to logspew in the polling so the app summary displays a log rate over zero.
* Also combined some tests to speed things up that were testing three things about the same output.
* Removed one SkipOnK8s call so we could merge tests.

N/A

None.

- [ ] YES
- [X] N/A

Should save about 2 minutes because we combined tests.

@rroberts2222

### DIFF
--- a/apps/lifecycle.go
+++ b/apps/lifecycle.go
@@ -246,14 +246,20 @@ var _ = AppsDescribe("Application Lifecycle", func() {
 				It("for memory usage", func() {
 					Eventually(func() float64 { m, _, _ := stats(); return m }, Config.CfPushTimeoutDuration()).Should(BeNumerically(">", 0.0))
 				})
+
 				Context("(cf-for-vms)", func() {
 					SkipOnK8s("App disk usage info unavailable")
 
 					It("for disk usage", func() {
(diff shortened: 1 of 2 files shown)


apps/lifecycle.go
```diff
@@ -243,18 +243,14 @@ var _ = AppsDescribe("Application Lifecycle", func() {
 					return mem, disk, logs
 				}
 
-				It("for memory usage", func() {
+				It("for memory, disk and log usage", func() {
 					Eventually(func() float64 { m, _, _ := stats(); return m }, Config.CfPushTimeoutDuration()).Should(BeNumerically(">", 0.0))
-				})
-				Context("(cf-for-vms)", func() {
-					SkipOnK8s("App disk usage info unavailable")
```
ctlong, line 250: Please add the `SkipOnK8s` back. Some folks who still maintain cf-for-k8s have asked to keep k8s-related things in CATs for the time being.

ctlong (changes requested):

ctlong (approved):

PR #650, opened 2022-11-29
Additional tests for r_buildpack and nginx_buildpack

jochenehret:

Additional tests for r_buildpack and nginx_buildpack, in particular to test the new cflinuxfs4 stack.

https://github.com/cloudfoundry/cf-deployment/issues/989

v23.3.0

- [x] introduces a new test
- [x] requires an update to a CATs integration-config (only if r_buildpack or nginx_buildpack don't have the standard names)

- [x] YES
- [ ] N/A

We want to verify that all buildpacks that come with cf-deployment support the new cflinuxfs4 stack. This is tested in the cf-deployment pipeline here:
https://app-runtime-deployments.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/experimental-cats-cflinuxfs4

Execution times were measured on https://app-runtime-deployments.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/experimental-cats

Before (229 specs): 39min 30sec average
After (331 specs): 37min

Looks like CATs on a warmed up CF run a bit faster... I would have expected a slight increase as we do 2 additional cf pushes.

### DIFF
--- a/detect/buildpacks.go
+++ b/detect/buildpacks.go
@@ -204,4 +204,40 @@ var _ = DetectDescribe("Buildpacks", func() {
 			})
 		}
 	})
+
+	Describe("nginx", func() {
+		for _, stack := range Config.GetStacks() {
+			Context(fmt.Sprintf("when using %s stack", stack), func() {
+				It("makes the app reachable via its bound route", func() {
+					Expect(cf.Cf("push", appName,
+						"-m", DEFAULT_MEMORY_LIMIT,
+						"-p", assets.NewAssets().Nginx,
+						"-s", stack,
+					).Wait(Config.DetectTimeoutDuration())).To(Exit(0))
+
+					Eventually(func() string {
+						return helpers.CurlAppRoot(Config, appName)
+					}).Should(ContainSubstring("Hello NGINX!"))
+				})
+			})
+		}
+	})
+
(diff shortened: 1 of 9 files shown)


detect/buildpacks.go
```diff
@@ -204,4 +204,40 @@ var _ = DetectDescribe("Buildpacks", func() {
 			})
 		}
 	})
+
+	Describe("nginx", func() {
+		for _, stack := range Config.GetStacks() {
```
ctlong, line 209: `stack` should be assigned to a local copy inside of the for loop, or else "the It closure will capture the mutating loop variable and all the specs will run against the last element in the loop." (re: [Dynamically Generating Specs](https://onsi.github.io/ginkgo/#dynamically-generating-specs))

ctlong: I see that most of our for loops on this page don't do this, and I think that they should all be changed while we're here.

ctlong (changes requested):

detect/buildpacks.go
jochenehret: Ah ok, I was scratching my head what `stack := stack` should be doing... will fix.

ctlong (approved):

PR #691, opened 2022-12-29
Fix: replace ioutil use as the package is deprecated

ctlong:

Yes

ioutil has been deprecated since Go 1.16 and should be replaced with io and os functions.

None

None

- [ ] YES
- [x] N/A

N/A

None

None


Benjamintf1 (approved):
looks good to me.

PR #693, opened 2023-01-04
Use configurable buildpack names instead of hard-coded strings

dsabeti:
Signed-off-by: Aram Price <pricear@vmware.com>

Some tests are still using hard-coded values to refer to buildpack names. These values can be retrieved from the Config object. Using the Config object in all places ensures that modifying the Config results in consistent behavior throughout CATs.

v24.6.0

- [ ] YES
- [x] N/A

N/A

None

@aramprice


ctlong (approved):
LGTM!

PR #694, opened 2023-01-05
Replace old java-spring asset with a new asset that also includes source code

dsabeti:
This will allow the asset to be rebuilt or updated as needed, while still providing a .jar file so that a Java toolchain is not needed to run CATs

Signed-off-by: David Sabeti <sabetid@vmware.com>

v24.6.0

- [ ] YES
- [x] N/A

N/A

N/A

@aramprice

### DIFF
--- a/assets/java-spring/README.md
+++ b/assets/java-spring/README.md
@@ -1,0 +1,19 @@
+# Spring Boot Trivial App
+
+This contains the built JAR, and the source code for a trivial SpringBoot app.
+
+The app has one feature which is to return `ok`.
+
+The source code is included so that future consumers of this app can make modifications.
+
+A pre-built JAR is included so that no Java toolchain is needed when running cats.
+
+## Building
+
+The application JAR can be rebuilt by running `mvn package` provided a JDK is
+available and Maven version  `3.5+` is in installed.
+
+## Context
+
+The app is build using SpringBoot 2.7.7, following the instructions found
+[here](https://docs.spring.io/spring-boot/docs/2.7.7/reference/html/getting-started.html#getting-started.first-application).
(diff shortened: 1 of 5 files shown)


ctlong (approved):

assets/java-spring/README.md
```diff
@@ -0,0 +1,19 @@
+# Spring Boot Trivial App
+
+This contains the built JAR, and the source code for a trivial SpringBoot app.
+
+The app has one feature which is to return `ok`.
```
aramprice, line 5: We opted to leave this behavior so that no tests required changes.

It would probably be helpful to future CATs operators / maintainers if this app returned something more meaningful, like `Hello from 'assets/java-spring/'` or something. Perhaps....

assets/java-spring/src/main/java/org/cloudfoundry/SpringBootTrivialApplication.java
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@EnableAutoConfiguration
+public class SpringBootTrivialApplication {
+    @RequestMapping("/")
+    String home() {
+        return "ok";
```
aramprice, line 13: See https://github.com/cloudfoundry/cf-acceptance-tests/pull/694/files#r1063008519

assets/java-spring/README.md
aramprice: See [[assets/java-spring/src/main/java/org/cloudfoundry/SpringBootTrivialApplication.java](https://github.com/cloudfoundry/cf-acceptance-tests/pull/694/files#diff-dee21702aae4a8ddb642ef9fbbf2cf8aabde12857406888312c4de2028b4fd17)](https://github.com/cloudfoundry/cf-acceptance-tests/pull/694/files#r1063008729) below

PR #697, opened 2023-01-10
Config test backfill

dsabeti:

As we were working on adding a new test suite, we found that there were some inconsistencies in the CATs Config tests. These commits attempt to address the following issues:
- De-duping struct definitions by embedding structs rather than duplicating values
- Rename the `allConfig` test object to `nullConfig` because it exists specifically to test null values in the config json.
  - We considered trying to write a test to ensure that the nullConfig struct always matches the Config struct to prevent this divergence from occurring again in the future. We decided it was out of scope for what we were hoping to achieve today.
- Added assertions for untested null value conditions
- Added assertions for overriding `include_*` config parameters as these were previously untested
- Added assertions for default config values for missing `include_*` config parameters

N/A (config test only)

- [ ] YES
- [x] N/A

N/A

N/A

@aramprice


dsabeti:
@ctlong Just a quick note on merging this: I'd recommend _not_ squashing these commits when you get to merging this. The change is simple, but verbose, and I think it would be hard to follow if all the changes came through in a single commit.

helpers/config/config_struct.go
```diff
@@ -686,6 +686,10 @@ func validateCredHubSettings(config *config) error {
 }
 
 func validateVolumeServices(config *config) error {
+	if config.IncludeVolumeServices == nil {
+		return nil
+	}
+
 	if !config.GetIncludeVolumeServices() {
```
ctlong, line 693: I don't think that we need a `nil` check for both `config.IncludeVolumeServices` and `config.GetIncludeVolumeServices()`, as they should be pulling the same value.

ctlong (commented):
@dsabeti @aramprice just one small comment. Otherwise it all LGTM!

Thanks!

helpers/config/config_struct.go
aramprice: I _think_ the issue is that `config.IncludeVolumeServices` is a pointer to a bool, so negating `config.GetIncludeVolumeServices()` will fail if the underlying pointer is `nil`.

The expression could probably be collapsed into something like:
`if config.GetIncludeVolumeServices() != nil && !config.GetIncludeVolumeServices() {`

ctlong: Ok, if that's the case, then it's fine as is.

PR #701, opened 2023-01-10
Remove python buildpack config

dsabeti:

Yes, however, this change does depend on first merging https://github.com/cloudfoundry/cf-acceptance-tests/pull/697.

 The `python_buildpack_name` field is never used by CATs, so this PR removes all references to it. In fact, despite having this configuration parameter, the Config _interface_ does not expose a `GetPythonBuildpackName()` function as it does with most other buildpacks, so we know programatically that this field is never used.

(The Python asset is used only in the `detect` test suite, where no buildpack name is specified on purpose -- the test exists to validate that a Python app is successfully detected by CF even when the user does not explicitly specify a buildpack.)

N/A

N/A

- [x] YES
- [ ] N/A

N/A

N/A

@aramprice

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,6 @@ include_v3
 * `nginx_buildpack_name` [See below](#buildpack-names)
 * `nodejs_buildpack_name` [See below](#buildpack-names)
 * `go_buildpack_name` [See below](#buildpack-names)
-* `python_buildpack_name` [See below](#buildpack-names)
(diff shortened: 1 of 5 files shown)


ctlong:
@dsabeti @aramprice I think this branch needs a rebase

ctlong:
@dsabeti @aramprice I think merging #700 created merge conflicts for this PR, and unfortunately another rebase is required.

ctlong (approved):

PR #747, opened 2023-02-22
Add support for positive + negative dynamic ASG tests

geofffranks:

Yes

Enables negative testing of dynamic asgs (ensuring that when disabled dynamic updates are not applied)

Based on the tests in 

Latest toolsmiths cf-deployment

Not sure if this requires an update to integration-config. If CATS is run against an environment with dynamic asgs disabled, and the security group testing enabled, you must also specify `dynamic_asgs_enabled: false` in your test config. When dynamic asgs is enabled (on by default), no changes are needed.

- [x] YES
- [ ] N/A

No new tests

Should not change the runtime for CATS under default circumstances. When dynamic asgs is enable, the test does not change. When disabled, it takes the same time as when enabled.

### DIFF
--- a/security_groups/dynamic_asgs.go
+++ b/security_groups/dynamic_asgs.go
@@ -50,7 +50,7 @@ var _ = Describe("Dynamic ASGs", func() {
 		deleteSecurityGroup(securityGroupName)
 	})
 
-	It("applies ASGs wihout app restart", func() {
+	It("applies ASGs without app restart", func() {
 		proxyRequestURL := fmt.Sprintf("%s%s.%s/https_proxy/cloud-controller-ng.service.cf.internal:9024/v2/info", Config.Protocol(), appName, Config.GetAppsDomain())
 
 		client := &http.Client{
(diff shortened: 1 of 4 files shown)


ctlong:
@geofffranks is there any way to dynamically ascertain whether or not dynamic ASGs are turned on?

geofffranks:
I don't think there's a way that CATS would have access to:

- parse the bosh manifest
- ssh onto a VM and look at configs/processes
- connect to locket and look for the presence of a lock

security_groups/dynamic_asgs.go
```diff
@@ -4,13 +4,15 @@ import (
 	"crypto/tls"
 	"fmt"
 	"io"
+	"io/ioutil"
```
ctlong, line 7: Can you please replace `ioutil` with the corresponding `io` or `os` functions. `ioutil` has been deprecated since go1.16.

security_groups/dynamic_asgs.go
```diff
@@ -4,13 +4,15 @@ import (
...
 	"net/http"
 	"time"
 
 	. "github.com/cloudfoundry/cf-acceptance-tests/cats_suite_helpers"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
```
ctlong, line 15: `gexec` is already dot-imported in this file.

ctlong (changes requested):

geofffranks:
@ctlong Thanks for catching those. I've pushed an update

ctlong (approved):

PR #752, opened 2023-02-24
Fix: remove validationerrors package in favor of go1.20 wrapped errors

ctlong:

Yes

As of go1.20, `errors.Join` provides a way to return an error wrapping a list of errors. This provides all the functionality that we need to do away with the `validationerrors` package.

We should use new code and clean up cruft.

https://tip.golang.org/doc/go1.20#errors

None

- [ ] YES
- [x] N/A

None

None

None


ctlong:
This shouldn't be undrafted until the repo's `go.mod` specifies go1.20 as the version for the project to use.

PR #753, opened 2023-02-24
Fix: error on incorrect timeout scale

ctlong:

Yes

Error out on invalid timeout scale rather than quietly correcting it to one. This is a breaking change.

I think this is useful to provide clarity to operators that their configuration is incorrect rather than silently choosing some value. If operators don't want to overwrite the timeout scale value, then they can remove the field from the JSON altogether, as there's a default value of 2.

None

v26.4.0

- [ ] YES
- [x] N/A

N/A

None

None

### DIFF
--- a/cats_suite_test.go
+++ b/cats_suite_test.go
@@ -47,97 +47,91 @@ import (
 const minCliVersion = "8.5.0"
 
 func TestCATS(t *testing.T) {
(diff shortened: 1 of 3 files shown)


ctlong:
Need to test this change against cf-d first.

davewalter (approved):

PR #817, opened 2023-03-28
fix: use curl error code to determine failure

staylor14:
* The version of `curl` that we get in cflinuxfs4 appears to have a
  different error message when the connection times out.
* The error code is consistant accross versions, so use it.

Authored-by: Steve Taylor <sttaylor@vmware.com>

Yes.

It appears that the `curl` that ships with cflinuxfs4 outputs a different string
when a connection fails.  The test is too specific, so we've added a match for
the curl error code (which has remained consistant).  The test now passes against
tas `rel/4.0`.

No further context available.

TAS 4.0.0 beta - we have not tested against cf-deployment yet, but the change is
additive, so we do not expect a difference in behavior for cf-deployment.

- [ ] YES
- [x] N/A

No meaningful change (sub milisecond)

#platform-provider-experience
@sttaylor
@dsabeti

### DIFF
--- a/tasks/task.go
+++ b/tasks/task.go
@@ -351,7 +351,7 @@ exit 1`
 					appLogs := logs.Recent(appName).Wait()
 					Expect(appLogs).To(Exit(0))
 					return string(appLogs.Out.Contents())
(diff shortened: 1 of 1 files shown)


dsabeti (approved):
Pretty straightforward, backwards-compatible addition to the regex

tjvman (approved):
LGTM

PR #833, opened 2023-04-18
asg timeout

ebroberson:

Yes

Change test for non-dynamic ASGs to use a sleep instead of Consistently because of intermittent connectivity issues.

While running `acceptance-tests` for `cf-networking-release` in ci, the `cats` section would error saying `connection refused`.  When we ran the tests manually we were unable to reproduce.  When repeating the test at 0.1s, we saw intermittent connection issues with ASGs. This new timeout should help prevent the connection error.

I believe v27.6.0.  It was a toolsmiths environment using the cf-deployment pool.

- [ ] YES
- [x] N/A

N/A

Hopefully no changes, but should make the test less flaky so we won't have to do multiple runs.

@reneighbor @geofffranks  @cloudfoundry/wg-app-runtime-interfaces

### DIFF
--- a/security_groups/dynamic_asgs.go
+++ b/security_groups/dynamic_asgs.go
@@ -15,11 +15,10 @@ import (
 
 	"github.com/cloudfoundry/cf-acceptance-tests/helpers/assets"
 	"github.com/cloudfoundry/cf-acceptance-tests/helpers/random_name"
-	"github.com/cloudfoundry/cf-acceptance-tests/helpers/skip_messages"
 	"github.com/cloudfoundry/cf-test-helpers/v2/cf"
 )
 
-var _ = Describe("Dynamic ASGs", func() {
+var _ = SecurityGroupsDescribe("ASGs", func() {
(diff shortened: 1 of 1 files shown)


geofffranks (dismissed):

ctlong:
@ebroberson I'm not sure that I'm understanding the issue correctly as described.. is it that the proxy app sometimes doesn't return the expected response when it's curled every 10ms?

If so, can you please just increase the interval of this specific `Consistently` to some duration that doesn't intermittently fail. You can do that by passing in a second argument (i.e. `Consistently(..., ...,500*time.Millisecond)`).

ebroberson:
Absolutely, I'll experiment and see if I can find an interval that works.

ebroberson:
@ctlong @dsabeti and I worked on this today and found that an Eventually looks like the best option to resolve this.  We also refactored the code a bit to make it easier to read.

ctlong (approved):

PR #838, opened 2023-05-02
v3 suite: use CfPushTimeotDuration where appropriate

dsabeti:

There are several tests that start an app, stage an app, or trigger CF to restart an app. In all of these cases, we should use the CfPushTimeoutDuration, because that duration is set aside specifically for events like these

N/A

- [ ] YES
- [x] N/A

N/A

This test changes some timeouts from 30 seconds to 240 seconds (or 4 minutes). There are 12 `Eventually`s changed in this commit. As such, in the worst case scenario, this could add up to (240-30)*12 = 2,520 seconds (or 42 minutes).

In all likelihood, this will not increase run time much for most consumers of CATs. In fact, it will only increase the runtime in cases where the tests were failing due to a timeout. So this is a trade between run time and robustness of the test suite to flakiness.

### DIFF
--- a/v3/buildpacks.go
+++ b/v3/buildpacks.go
@@ -172,7 +172,7 @@ var _ = V3Describe("buildpack", func() {
 
 			Eventually(func() string {
 				return helpers.CurlAppRoot(Config, webProcess.Name)
-			}).Should(ContainSubstring("The bundler version is"))
+			}, Config.CfPushTimeoutDuration()).Should(ContainSubstring("The bundler version is"))
 		})
 	})
 })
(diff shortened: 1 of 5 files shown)


ctlong (approved):
LGTM.

@dsabeti are you seeing CATs failures due to insufficient timeouts for these tests, just curious?

Technically the `CfPushTimeoutDuration` is the timeout for `cf push` commands to succeed, so I assume that wasn't added here because these cases don't directly use that command. Seems perfectly valid to include them though, and I prefer that to a new timeout parameter if the current default isn't cutting it.

PR #902, opened 2023-07-31
Rework cats pipeline

jochenehret:

As discussed in the ARD WG meeting from Jul 27th, we want to rework the CATs release pipeline:

* use cats-develop branch as main trigger
* first run unit tests as sanity check
* then deploy cf from cf-deployment "main" branch
* run smoke-tests and then the actual CATs tests
* finally, cleanup CATs and destroy cf again to save cost

Note that the smoke-tests pipeline already uses a separate bbl infrastructure (was shared with CATs before).

As the pipeline.yml diff is likely not easy to read, I've uploaded the new pipeline (in paused state) here:
https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cats-new

(not relevant)

- [ ] YES
- [x] N/A

### DIFF
--- a/ci/cats.md
+++ b/ci/cats.md
@@ -1,4 +1,4 @@
-# CATs
+# cats
 
 ## Purpose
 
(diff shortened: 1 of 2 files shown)


ctlong (dismissed):
LGTM

jochenehret:
Thanks for the quick reply, I would leave this PR open (and the new pipeline paused) for one more day to give more people the chance for a review.

ctlong (approved):

Issue #862, opened 2023-06-07
Zipkin test failures


ctlong:
We've been seeing a lot of flakiness in the CATs [zipkin test](https://github.com/cloudfoundry/cf-acceptance-tests/blob/d75365723921c060502ab422dbc39382331c2d0c/routing/zipkin_tracing.go) recently. Particularly when trying to push the [spring-sleuth asset](https://github.com/cloudfoundry/cf-acceptance-tests/tree/4ae853836e0bda3972ab58e89a09aa6f3b824f25/assets/spring-sleuth), which happens to be ~7 years old.

```
Summarizing 1 Failure:
  [FAIL] [routing] Zipkin Tracing [BeforeEach] when zipkin tracing is enabled when zipkin headers are not in the request the sleuth error response has no error
  /go/src/github.com/cloudfoundry/cf-acceptance-tests/routing/zipkin_tracing.go:36
```

<details><summary>App Report</summary>
<p>

```
***************************************************
  ***** APP REPORT: CATS-7-APP-9bfe688507c8088a *****
  ***************************************************

  [2023-06-07 05:08:48.93 (UTC)]> cf app CATS-7-APP-9bfe688507c8088a --guid 
  64db5f4d-a109-4fb2-b052-1990926a309c

  [2023-06-07 05:08:49.93 (UTC)]> cf logs --recent CATS-7-APP-9bfe688507c8088a 
  Retrieving logs for app CATS-7-APP-9bfe688507c8088a in org CATS-7-ORG-d1655112e09a7151 / space CATS-7-SPACE-713a9c3fc1938cde as CATS-7-USER-a9f1e7301fe48695...

     2023-06-07T05:03:52.83+0000 [API/1] OUT Added process: "web"
     2023-06-07T05:03:52.84+0000 [API/1] OUT Created app with guid 64db5f4d-a109-4fb2-b052-1990926a309c
     2023-06-07T05:03:52.86+0000 [API/1] OUT Applied manifest to app with guid 64db5f4d-a109-4fb2-b052-1990926a309c (---
     2023-06-07T05:03:52.87+0000 [API/1] OUT applications:
     2023-06-07T05:03:52.87+0000 [API/1] OUT - name: CATS-7-APP-9bfe688507c8088a
     2023-06-07T05:03:52.87+0000 [API/1] OUT path: "/go/src/github.com/cloudfoundry/cf-acceptance-tests/assets/spring-sleuth/spring-sleuth.jar"
[...]

PR #914, opened 2023-08-09
fix(zipkin): increase app readiness timeout to reduce test flakes

ctlong:

Increase readiness timeout for Spring Sleuth app in Zipkin Tracking tests. This should hopefully reduce test flakes in this suite.

Also refactors the zipkin tests, just cause.

#862 

v29.1.0

- [ ] YES
- [x] N/A

N/A

None

None


johha (approved):
lgtm 👍

PR #915, opened 2023-08-09
new cat for readiness health check feature - linux & windows

ameowlia:
✅ Yes.

This PR adds a new CAT for readiness health checks, a feature that is coming out with new versions of CAPI and Diego soon.

* [CFF RFC for adding readiness health checks](https://github.com/cloudfoundry/community/blob/rfc-readiness-healthchecks/toc/rfc/rfc-draft-readiness-healthchecks.md)
* [PR to add this to Cloud Controller](https://github.com/cloudfoundry/cloud_controller_ng/pull/3351)

⛔️⛔️⛔️TBD⛔️⛔️⛔️

- [x] YES
- [ ] N/A

We are adding a major new feature to CF. This CAT will test one basic workflow for that feature.

220-ish seconds. Sorry :( 

* @mariash


apps/readiness_healthcheck.go
```diff
@@ -0,0 +1,134 @@
...
+		app_helpers.AppReport(appName)
+		app_helpers.AppReport(proxyAppName)
+		Eventually(cf.Cf("delete", appName, "-f")).Should(Exit(0))
+		Eventually(cf.Cf("delete", proxyAppName, "-f")).Should(Exit(0))
+	})
+
+	Describe("when the readiness healthcheck is set to http", func() {
+		FIt("registers the route only when the readiness check passes", func() {
+			By("pushing the app")
+			Expect(cf.Push(appName,
```
ctlong, line 47: You should use `cf.Cf("push",...)` instead. That way no changes are required to [cloudfoundry/cf-test-helpers](https://github.com/cloudfoundry/cf-test-helpers).

ameowlia: 💡 Good idea! @geofffranks you should do that :)

geofffranks: Done

windows/readiness_healthcheck.go
```diff
@@ -0,0 +1,98 @@
...
+			By("verifying the app is marked as not ready")
+
+			// TODO: only include this when audit events are built
+			// Eventually(func() string {
+			// 	return string(cf.Cf("events", appName).Wait().Out.Contents())
+			// }).Should(MatchRegexp("app.notready"))
+
+			Eventually(func() string {
+				return string(logs.Recent(appName).Wait().Out.Contents())
+			}, readinessHealthCheckTimeout).Should(ContainSubstring("Container failed the readiness health check"))
```
ctlong, line 71: Please refactor with gbytes

windows/readiness_healthcheck.go
```diff
@@ -0,0 +1,98 @@
...
+
+			By("verifying the app is removed from the routing table")
+			Eventually(func() string {
```
ctlong, line 74: Should this be `Consistently` instead of `Eventually`?

apps/readiness_healthcheck.go
```diff
@@ -0,0 +1,133 @@
...
+			// Eventually(func() string {
+			// 	return string(cf.Cf("events", appName).Wait().Out.Contents())
+			// }).Should(MatchRegexp("app.notready"))
+
+			Eventually(func() string {
+				return string(logs.Recent(appName).Wait().Out.Contents())
+			}, readinessHealthCheckTimeout).Should(ContainSubstring("Container failed the readiness health check"))
+
+			By("verifying the app is removed from the routing table")
+			Eventually(func() string {
```
ctlong, line 88: Should this be `Consistently` instead of `Eventually`?

windows/readiness_healthcheck.go
```diff
@@ -0,0 +1,98 @@
...
+				return helpers.CurlApp(Config, appName, "/ready")
+			}).Should(ContainSubstring("404 Not Found"))
+
+			By("verifying that the app hasn't restarted")
+			Consistently(func() string {
+				return string(cf.Cf("events", appName).Wait().Out.Contents())
+			}).ShouldNot(MatchRegexp("audit.app.process.crash"))
+
+			By("re-enabling the app's readiness endpoint")
+			Expect(cf.Cf("ssh", appName, "-c", "curl localhost:8080/ready/true").Wait(Config.DefaultTimeoutDuration())).To(Exit(0))
```
ctlong, line 84: Is there another way to re-enable the app's readiness endpoint? The Diego container ssh feature is not guaranteed to be turned on for this suite.

ctlong (commented):
I've left a bunch of comments for your review. Generally it looks good.

🤔 I do wonder whether there's a way we can split up the tests to reduce the time this adds to CATs. I'm thinking that maybe the test could be split in two:
1. Push an app that will fail the readiness health check (i.e. push with `--no-start` and set an env var that will cause readiness to health check to fail at first), then reactivate it with the proxy app c2c and see everything work.
2. Push an app that will pass the readiness health check, then deactivate the readiness health check with the endpoint, and see it removed from the routing chain.

windows/readiness_healthcheck.go
geofffranks: Unfortunately no. The method used for linux is to use c2c networking (which isnt' eaven a feature on windows). 
we can remove the re-enable checks if it sounds like this will be a large concern.

geofffranks: or maybe there's a boolean for app ssh we can hide it behind?

apps/readiness_healthcheck.go
geofffranks: No, this was eventually to ensure that the delay between catching the readiness failure + route-emitter updating nats + gorouter. 

Arguably we could have another consistently check for the same thing after, but that would likely need to wait 40s+.

windows/readiness_healthcheck.go
geofffranks: No, this was eventually to ensure that the delay between catching the readiness failure + route-emitter updating nats + gorouter. 

Arguably we could have another consistently check for the same thing after, but that would likely need to wait 40s+.

windows/readiness_healthcheck.go
```diff
@@ -0,0 +1,98 @@
...
+				return helpers.CurlAppRoot(Config, appName)
+			}, Config.CfPushTimeoutDuration()).Should(ContainSubstring("hello i am nora running on"))
+
+			By("verifying the app is marked as ready")
+			// TODO: only include this when audit events are built
+			// Eventually(func() string {
+			// 	return string(cf.Cf("events", appName).Wait().Out.Contents())
+			// }).Should(MatchRegexp("app.ready"))
+
+			// Windows apps take a Loooooong time to start apparently
```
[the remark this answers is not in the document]
geofffranks: Possibly, will investigate here.

apps/readiness_healthcheck.go
```diff
@@ -0,0 +1,133 @@
...
+		Eventually(cf.Cf("delete", proxyAppName, "-f")).Should(Exit(0))
+	})
+
+	Describe("when the readiness healthcheck is set to http", func() {
+		It("registers the route only when the readiness check passes", func() {
+			By("pushing the app")
+
+			Expect(cf.Cf("push", appName,
+				"-p", assets.NewAssets().Dora,
+				"-f", assets.NewAssets().Dora+"/readiness_manifest.yml",
```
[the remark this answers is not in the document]
geofffranks: The flags are not present in CF CLI  yet, and won't be for some time after the feature is released in CAPI.

.gitignore
```diff
@@ -12,3 +12,5 @@ assets/dora.zip
 assets/credhub-enabled-app/build
 assets/credhub-enabled-app/.gradle
 bin/ginkgo
+/assets/nora/.vs/Nora
+/.vs
```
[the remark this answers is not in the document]
geofffranks: Will do, iirc this is what vscode added for the gitignore on its own 😂

apps/readiness_healthcheck.go
ctlong: I guess I expected the "Container failed the readiness health check" log message, or some comparable log message, to coincide with the app being removed from the routing table. Then after receiving that, using `Consistently` would make sense to check that the routing consistently doesn't work.

windows/readiness_healthcheck.go
ctlong: I guess I expected the "Container failed the readiness health check" log message, or some comparable log message, to coincide with the app being removed from the routing table. Then after receiving that, using `Consistently` would make sense to check that the routing consistently doesn't work.

windows/readiness_healthcheck.go
ctlong: Hmmmm ok in that case let's roll with this for now. In the worst-case, I can fast-follow this major release with a patch to gate this test on the `include_ssh` config option.

windows/readiness_healthcheck.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		appName = random_name.CATSRandomName("APP")
+	})
+
+	AfterEach(func() {
+		app_helpers.AppReport(appName)
+		Eventually(cf.Cf("delete", appName, "-f")).Should(Exit(0))
+	})
+
+	// TODO add a ready endpoint to Nora
```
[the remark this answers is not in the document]
geofffranks: 👍

windows/readiness_healthcheck.go
```diff
@@ -0,0 +1,98 @@
...
+	Describe("when the readiness healthcheck is set to http", func() {
+		It("registers the route only when the readiness check passes", func() {
+			By("pushing the app")
+			// FIXME use manifest and update linux
```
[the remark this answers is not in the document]
geofffranks: 👍

windows/readiness_healthcheck.go
```diff
@@ -0,0 +1,98 @@
...
+			}, Config.CfPushTimeoutDuration()).Should(ContainSubstring("hello i am nora running on"))
+
+			By("verifying the app is marked as ready")
+			// TODO: only include this when audit events are built
+			// Eventually(func() string {
+			// 	return string(cf.Cf("events", appName).Wait().Out.Contents())
+			// }).Should(MatchRegexp("app.ready"))
+
+			// Windows apps take a Loooooong time to start apparently
+			Expect(string(logs.Recent(appName).Wait(Config.CfPushTimeoutDuration()).Out.Contents())).Should(ContainSubstring("Container passed the readiness health check"))
```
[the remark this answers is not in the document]
geofffranks: 👍

windows/readiness_healthcheck.go
geofffranks: 👍

vendor/github.com/cloudfoundry/cf-test-helpers/v2/cf/push.go
```diff
@@ -17,13 +17,16 @@ type Manifest struct {
...
-	Buildpacks []string `yaml:",omitempty"`
-	Command    string   `yaml:",omitempty"`
-	Instances  int      `yaml:",omitempty"`
-	Memory     string   `yaml:",omitempty"`
-	Name       string
-	Path       string              `yaml:",omitempty"`
-	Routes     []map[string]string `yaml:",omitempty"`
+	Buildpacks                       []string `yaml:",omitempty"`
+	Command                          string   `yaml:",omitempty"`
+	Instances                        int      `yaml:",omitempty"`
```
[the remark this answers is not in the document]
geofffranks: 👍

assets/nora/Nora.Tests/obj/Release/Nora.Tests.csproj.FileListAbsolute.txt
```diff
@@ -0,0 +1,20 @@
...
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora.Tests\bin\Release\Nora.dll
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora.Tests\bin\Release\NSpec.dll
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora.Tests\bin\Release\System.Net.Http.Formatting.dll
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora.Tests\bin\Release\System.Web.Http.dll
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora.Tests\bin\Release\MySql.Data.dll
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora.Tests\bin\Release\System.Web.Http.WebHost.dll
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora.Tests\bin\Release\Nora.pdb
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora.Tests\bin\Release\System.Net.Http.Formatting.xml
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora.Tests\bin\Release\System.Web.Http.xml
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora.Tests\bin\Release\System.Web.Http.WebHost.xml
```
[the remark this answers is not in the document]
geofffranks: Nope. removed + gitignore updated.

assets/nora/Nora/obj/Release/Nora.csproj.FileListAbsolute.txt
```diff
@@ -0,0 +1,16 @@
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora\bin\Nora.dll
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora\bin\Nora.pdb
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora\bin\MySql.Data.dll
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora\bin\Newtonsoft.Json.dll
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora\bin\System.Net.Http.Formatting.dll
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora\bin\System.Web.Http.dll
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora\bin\System.Web.Http.WebHost.dll
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora\bin\System.Net.Http.Formatting.xml
+C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora\bin\System.Web.Http.xml
```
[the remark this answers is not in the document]
geofffranks: Nope. removed + gitignore updated.

apps/readiness_healthcheck.go
```diff
@@ -0,0 +1,133 @@
...
+			By("verifying the app is marked as not ready")
+
+			// TODO: only include this when audit events are built
+			// Eventually(func() string {
+			// 	return string(cf.Cf("events", appName).Wait().Out.Contents())
+			// }).Should(MatchRegexp("app.notready"))
+
+			Eventually(func() string {
+				return string(logs.Recent(appName).Wait().Out.Contents())
+			}, readinessHealthCheckTimeout).Should(ContainSubstring("Container failed the readiness health check"))
```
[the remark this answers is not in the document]
geofffranks: 👍

apps/readiness_healthcheck.go
```diff
@@ -0,0 +1,133 @@
...
+			Eventually(func() string {
+				return helpers.CurlApp(Config, appName, "/ready")
+			}, Config.DefaultTimeoutDuration()).Should(ContainSubstring("200 - ready"))
+
+			// TODO: only include this when audit events are built
+			// Eventually(func() string {
+			// 	return string(cf.Cf("events", appName).Wait().Out.Contents())
+			// }).Should(MatchRegexp("app.ready"))
+
+			Expect(string(logs.Recent(appName).Wait().Out.Contents())).Should(ContainSubstring("Container passed the readiness health check"))
```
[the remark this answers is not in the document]
geofffranks: 👍

apps/readiness_healthcheck.go
ameowlia: > Container failed the readiness health check" log message, or some comparable log message, to coincide with the app being removed from the routing table. 

Sadly, this is not true, but I can see how this would be the assumption, and it is definitely confusing.

It looks something like this....
1. The readiness check gets a non-2xx. Marks the app as not-ready. Logs the not-ready bit in the logs.
2. Eventually (every 20s), the route emitter syncs, sees that the app is not ready and removes it from the routing table.

That's why it is Eventually and not Consistently.

geofffranks:
Feedback has been addressed/commented on. I looked into splitting them out into multiple tests to parallelize. Unfortunately we don't have any feedback from an app starting up that it has failed the initial readiness checks, other than absence of a message. Short of waiting for a while, we can't differentiate "readiness hasn't checked yet" and "readiness ran and failed." This makes the second test run slightly longer than when a single test runs both cases.

ctlong:
> I looked into splitting them out into multiple tests to parallelize... This makes the second test run slightly longer than when a single test runs both cases.

@geofffranks and I chatted about this and discovered that he accidentally ran his comparison with no parallelization, as `./bin/test` does not run CATs in parallel by default. The `--procs` flag needs to be explicitly provided.

👉 I'm waiting for that comparison to be re-ran with parallelization before I review this PR again. 👈

geofffranks:
I re-ran with `--procs 8` and as a single test, it executed in 104s. As a split test, it took 148s.

The main difference is that when split, the un-ready app has to wait in a Consistently() for 25s to ensure the app never becomes ready, which the original test doesn't have.  

Going to leave this as a single test.

assets/nora/Nora.Tests/obj/Release/Nora.Tests.csproj.FileListAbsolute.txt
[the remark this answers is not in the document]
ctlong: @geofffranks this is still part of the PR. You might need to un-gitignore these files, remove them, then gitignore them again.

assets/nora/Nora/Properties/PublishProfiles/FolderProfile.pubxml
```diff
@@ -0,0 +1,17 @@
...
+-->
+<Project>
+  <PropertyGroup>
+    <DeleteExistingFiles>false</DeleteExistingFiles>
+    <ExcludeApp_Data>false</ExcludeApp_Data>
+    <LaunchSiteAfterPublish>true</LaunchSiteAfterPublish>
+    <LastUsedBuildConfiguration>Release</LastUsedBuildConfiguration>
+    <LastUsedPlatform>Any CPU</LastUsedPlatform>
+    <PublishProvider>FileSystem</PublishProvider>
+    <PublishUrl>C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\NoraPublished</PublishUrl>
```
ctlong, line 13: Here's another example of a path that we might not want to commit

assets/nora/Nora/Properties/PublishProfiles/FolderProfile.pubxml.user
```diff
@@ -0,0 +1,100 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!--
+https://go.microsoft.com/fwlink/?LinkID=208121.
+-->
+<Project>
+  <PropertyGroup>
+    <_PublishTargetUrl>C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\NoraPublished</_PublishTargetUrl>
```
ctlong, line 7: Here's another example of a path that we might not want to commit

assets/nora/NoraPublished/Nora.csproj.user
```diff
@@ -0,0 +1,39 @@
...
+  <PropertyGroup>
+    <LastActiveSolutionConfig>Release|Any CPU</LastActiveSolutionConfig>
+    <UseIISExpress>true</UseIISExpress>
+    <Use64BitIISExpress />
+    <IISExpressSSLPort />
+    <IISExpressAnonymousAuthentication />
+    <IISExpressWindowsAuthentication />
+    <IISExpressUseClassicPipelineMode />
+    <UseGlobalApplicationHostFile />
+    <NameOfLastUsedPublishProfile>C:\Users\gfranks\source\repos\cf-acceptance-tests\assets\nora\Nora\Properties\PublishProfiles\FolderProfile.pubxml</NameOfLastUsedPublishProfile>
```
ctlong, line 12: Here's another example of a path that we might not want to commit

ctlong:
I'm still seeing the local path `C:\Users\gfranks\source\repos\cf-acceptance-tests` referenced throughout the changes to Nora. I don't know .NET well enough to know when it's avoidable or not, but hopefully most are 🤞 

Can you all please search for the term `gfranks` in the `assets/nora` path (make sure to check hidden and ignored files!) and remove them wherever possible.

ctlong (approved):

ctlong:
Aaaah shoot, can't believe I didn't catch this earlier 🤦 

The base branch for this PR needs to be `develop`. Can you please change the base branch and rebase.

ctlong (approved):

go.mod
```diff
@@ -39,6 +39,7 @@ require (
 	github.com/ulikunitz/xz v0.5.10 // indirect
 	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 // indirect
 	golang.org/x/net v0.12.0 // indirect
+
```
ctlong, line 42: This probably shouldn't be here. Can you run the following:
```
go mod tidy && go mod vendor
```

geofffranks: Fixed

ctlong (approved):

ctlong (approved):

PR #1068, opened 2024-02-27
Add overall test timeout setting

kei-yamazaki:

From Ginkgo 1.0 to 2.0, the default timeout for the entire test changed from `24h` to `1h` (refer to: [Ginkgo 2.0 Migration Guide - Timeout Behavior](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#timeout-behavior)).

Depending on the environment in which the test is run, the execution time of the test may exceed 1h, which may cause the test to fail.

Therefore, I have added an explanation regarding the timeout setting for the entire test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,6 +330,19 @@ To see verbose output from `ginkgo`, use the `-v` flag.
 
 You can of course combine the `-v` flag with the `--procs=N` flag.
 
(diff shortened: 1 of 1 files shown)


ctlong (approved):
Thanks!

PR #1106, opened 2024-04-05
Correctly print CF CLI version

johha:
STDOUT output of `fmt` is not captured by ginkgo and thus the version is not printed. When using `GinkgoWriter` the message is captured correctly.

-> not  applicable

- [ ] YES
- [x] N/A


jochenehret (approved):

ctlong:
@johha do we still need the `PauseOutputInterception()` and `ResumeOutputInterception()` calls around it now?

johha:
> @johha do we still need the `PauseOutputInterception()` and `ResumeOutputInterception()` calls around it now?

Still need it to show the version number

PR #1136, opened 2024-05-14
Add acceptance tests for cnb lifecycle

c0d1ngm0nk3y:

Added acceptance tests using the new cloud native buildpack lifecycle.

* RFC for cloud native buildpacks - https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0028-cnb-lifecycle.md
* Prerequisite 1 - Changes in [cloud_controller_ng](https://github.com/cloudfoundry/cloud_controller_ng/pull/3778)
* Prerequisite 2 - Changes in [diego-release](https://github.com/cloudfoundry/diego-release/pull/929)
* Prerequisite 3 - Changes in [cf-deployment](https://github.com/cloudfoundry/cf-deployment/pull/1178)

* A future version - see prerequisite 3

- [X] introduces a new test --- Are you sure everyone should be running this test?

- [X] YES
- [ ] N/A

Same as for docker or buildpacks lifecycle. It is some basic functionality added to `cf`.

~ 70s

@loewenstein-sap, @modulo11, @pbusko, @nicolasbender

### DIFF
--- a/cnb/cnb_lifecycle.go
+++ b/cnb/cnb_lifecycle.go
@@ -1,6 +1,3 @@
-//go:build !noInternet && !noCNB
-// +build !noInternet,!noCNB
-
 package cnb
 
 import (
--- a/cnb/cnb_lifecycle.go
+++ b/cnb/cnb_lifecycle.go
@@ -1,0 +1,67 @@
+//go:build !noInternet && !noCNB
+// +build !noInternet,!noCNB
+
+package cnb
+
+import (
+	"path/filepath"
+
+	. "github.com/cloudfoundry/cf-acceptance-tests/cats_suite_helpers"
+	"github.com/cloudfoundry/cf-acceptance-tests/helpers/app_helpers"
+	"github.com/cloudfoundry/cf-acceptance-tests/helpers/assets"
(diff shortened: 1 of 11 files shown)


cnb/cnb_lifecycle.go
```diff
@@ -0,0 +1,67 @@
+//go:build !noInternet && !noCNB
+// +build !noInternet,!noCNB
```
ctlong, line 2: Why add these build tags? I did a search and see that there are a couple similar build tags, but I don't see them justified anywhere. Seems like we should lean into using our configuration parameters to skip tests we don't want to run.

ctlong (changes requested):
The test LGTM!

Just a few documentation requests before I think this is ready for review.

cnb/cnb_lifecycle.go
modulo11: We started with a copy of the docker lifecycle tests. If those tags are not used/mandatory, we can just remove them.

ctlong: Ok, AFAIK they're unused. If you could remove them please, that would be great.

ctlong: Oh seems like you did, thanks.

Issue #1173, opened 2024-07-11
Flaky test: [tcp routing] TCP Routing external ports with a second external port [It] maps both ports to the same application


jochenehret:
The TCP Routing test that checks if one app can be reached from two ports is failing often here:
https://github.com/cloudfoundry/cf-acceptance-tests/blob/6f060209f7a55f0c4f8d0fffabb122c785ce914e/tcp_routing/tcp_routing.go#L131

Example failures:
https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/fips-cats/builds/82
https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/fips-cats/builds/57
https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/fips-cats/builds/113
https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/fips-cats/builds/120

I've recreated the test setup manually on fips/snape. The setup works as expected: You can send data over two different TCP ports to the test app and the app responds as expected. Running the test in the CATs suite however fails often.

I've added some debug statements with timestamps. Here's the flow from a failed run:

```
# sending first test message to first port
# https://github.com/cloudfoundry/cf-acceptance-tests/blob/6f060209f7a55f0c4f8d0fffabb122c785ce914e/cats_suite_helpers/cats_suite_helpers.go#L406
starting SendAndReceive(tcp.cf.snape.env.wg-ard.ci.cloudfoundry.org, 1031) at Jul 11 14:49:45.862

# output from test app: https://github.com/cloudfoundry/cf-acceptance-tests/blob/6f060209f7a55f0c4f8d0fffabb122c785ce914e/assets/tcp-listener/main.go#L53
# "10.0.32.11" is one of the two tcp-routers
2024-07-11T12:49:45.97+0000 [APP/PROC/WEB/0] OUT Message to 10.0.32.11:41084: server1:Time is 938260798
2024-07-11T12:49:45.99+0000 [APP/PROC/WEB/0] OUT Jul 11 14:49:45.991 (read) Closing connection to 10.0.32.11:41084: EOF

# sending second test message to other port
starting SendAndReceive(tcp.cf.snape.env.wg-ard.ci.cloudfoundry.org, 1026) at Jul 11 14:49:45.955

[...]

PR #1174, opened 2024-07-11
[Issue 1173] In "SendAndReceive", wait 100ms before reading

jochenehret:

Prevents race condition when sending a test message to a backend and reading the response.

https://github.com/cloudfoundry/cf-acceptance-tests/issues/1173

v41.0.0

- [ ] YES
- [x] N/A

Function is called 7 times -> 700ms.

### DIFF
--- a/cats_suite_helpers/cats_suite_helpers.go
+++ b/cats_suite_helpers/cats_suite_helpers.go
@@ -415,7 +415,7 @@ func SendAndReceive(addr string, externalPort string) (string, error) {
 	}
 
 	// see https://github.com/cloudfoundry/cf-acceptance-tests/issues/1173
-	time.Sleep(2 * time.Second)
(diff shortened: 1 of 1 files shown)


ctlong (commented):
Could we resolve the problem by using an `Eventually` when calling `SendAndReceive` instead? I would prefer that approach in order to allow the test to succeed faster, if possible.

jochenehret:
Using `Eventually` when calling `SendAndReceive` could also fix the problem. But logically, I think it makes sense to add a short wait before the `Read` call. The tests are running on our K8s Concourse on GCP. The test apps runs on an AWS account behind a network load balancer. So the full roundtrip is expected to take a short amount of time. I would propose to wait 100ms and see how that works.

ctlong (approved):

PR #1190, opened 2024-08-13
Use 1.22 as the go directive instead of 1.22.0

ctlong:

Yes

Updates assets and base go.mod to use 1.22 as the go directive instead of 1.22.0. Some tooling in buildpacks depends upon the go directive being only in <major>.<minor> format. And it seems like the base go.mod should match.

https://github.com/cloudfoundry/cf-acceptance-tests/commit/043a63ff66b5a67c36a14876edf9095d74b62783#commitcomment-145272348

None

- [ ] YES
- [x] N/A

N/A

None

None

### DIFF
--- a/assets/binary/go.mod
+++ b/assets/binary/go.mod
@@ -1,3 +1,3 @@
 module github.com/cloudfoundry/cf-acceptance-tests/assets/binary
 
-go 1.22.0
+go 1.22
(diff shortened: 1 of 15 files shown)


dimivel (approved):
LGTM

davewalter (approved):

PR #1195, opened 2024-08-16
Set go toolchain explicitly

Samze:

It seems like theres an issue with using `go 1.22.0` with buildpacks as per this commit https://github.com/cloudfoundry/cf-acceptance-tests/commit/a0e5ea36d59c6d7ba56767391ab05bcabe2ca81b .However, the directive `go 1.22` this is not a valid toolchain and so builind the app will result in
```
$ pwd
/Users/sgunaratne/workspace/cf-acceptance-tests/assets/worker
$ go mod tidy
go: downloading go1.22 (darwin/arm64)
go: download go1.22 for darwin/arm64: toolchain not available
```

This adds the toolchain explicitly so work around this issue.

```
$ git diff go.mod
diff --git a/assets/worker/go.mod b/assets/worker/go.mod
index d1180cc5..d3a08650 100644
--- a/assets/worker/go.mod
+++ b/assets/worker/go.mod
@@ -1,3 +1,5 @@
 module github.com/cloudfoundry/cf-acceptance-tests/assets/worker

 go 1.22
+
+toolchain go1.22.0
$ go mod tidy
$ echo $?
0
```

See https://github.com/golang/go/issues/62278#issuecomment-1693538776 for more context on why this is required in more recent go versions.

This is failing in the cf-cli pipeline https://github.com/cloudfoundry/cli/actions/runs/10423472162/job/28875001824?pr=3113
```
     2024-08-16T20:01:58.74+0000 [STG/0] OUT **WARNING** go 1.21.x will no longer be available in new buildpacks released after 2024-08-15.
     2024-08-16T20:01:58.74+0000 [STG/0] OUT See: https://golang.org/doc/devel/release.html
     2024-08-16T20:02:03.66+0000 [STG/0] OUT **ERROR** problem retrieving main package name: go: downloading go1.22 (linux/amd64)
     2024-08-16T20:02:03.66+0000 [STG/0] OUT go: download go1.22 for linux/amd64: toolchain not available
     2024-08-16T20:02:03.66+0000 [STG/0] OUT 
     2024-08-16T20:02:03.66+0000 [STG/0] OUT **ERROR** Unable to determine import path: exit status 1
     2024-08-16T20:02:04.48+0000 [STG/0] ERR Failed to compile droplet: Failed to run finalize script: exit status 12
```

- [ ] YES
- [x] N/A

### DIFF
--- a/assets/binary/go.mod
+++ b/assets/binary/go.mod
@@ -1,3 +1,5 @@
 module github.com/cloudfoundry/cf-acceptance-tests/assets/binary
 
 go 1.22
+
+toolchain go1.22.0
--- a/assets/catnip/go.mod
+++ b/assets/catnip/go.mod
@@ -2,6 +2,8 @@ module github.com/cloudfoundry/cf-acceptance-tests/assets/catnip
 
 go 1.22
 
+toolchain go1.22.0
+
 require (
 	code.cloudfoundry.org/clock v1.2.0
 	github.com/go-chi/chi/v5 v5.1.0
--- a/assets/credhub-service-broker/go.mod
+++ b/assets/credhub-service-broker/go.mod
@@ -2,6 +2,8 @@ module github.com/cloudfoundry/cf-acceptance-tests/assets/credhub-service-broker
 
 go 1.22
 
+toolchain go1.22.0
+
 require (
 	code.cloudfoundry.org/credhub-cli v0.0.0-20230320130818-a7d5420b283b
 	github.com/go-chi/chi/v5 v5.1.0
--- a/assets/go_calls_ruby/go.mod
+++ b/assets/go_calls_ruby/go.mod
@@ -1,3 +1,5 @@
 module github.com/cloudfoundry/cf-acceptance-tests/assets/go_calls_ruby
 
 go 1.22
+
+toolchain go1.22.0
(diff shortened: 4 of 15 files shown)


ctlong:
I was afraid of something like this.. I was aware of https://github.com/golang/go/issues/62278#issuecomment-1693538776, but set the go module versions to `1.22` instead of `1.22.0` anyway (in #1190) in order to fix [this](https://github.com/cloudfoundry/cf-acceptance-tests/commit/043a63ff66b5a67c36a14876edf9095d74b62783#commitcomment-145358993) user problem.

ctlong:
I'm wary of using `toolchain 1.22.0` as that doesn't represent the "suggested Go toolchain to use with a module." I guess this might be the way to go though, as a local user would likely have a GOTOOLCHAIN env var set that overrides the toolchain to be more recently.

a-b:
@ctlong  - if @Samze's proposal causes new challenges, what would be the alternative or acceptable approach?

ctlong:
I think this probably is the way to fix this problem. The only other thing that comes to mind is reverting the change to `1.22` from `1.22.0`, but since that will likely break other people let's try this first.

ctlong (approved):

PR #1203, opened 2024-08-23
Update CF_INSTANCE_PORT and CF_INSTANCE_ADDR tests so they pass regar…

geofffranks:

yes

Adjusts env var tests related to CF_INSTANCE_PORT and CF_INSTANCE_ADDR to pass regardless of how unproxied ports are configured for app containers.

Allows https://github.com/cloudfoundry/cf-deployment/pull/1195 to no longer need changes to CATs  configs

Latest.

- [x] YES
- [ ] N/A

N/A

no change

- [] **Urgent** - unblocks current or future work

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,8 +161,6 @@ include_app_syslog_tcp
 * `private_docker_registry_password`: Password to access the private docker repository. [See below](#private-docker)
 * `unallocated_ip_for_security_group`: An unused IP address in the private network used by CF. Defaults to 10.0.244.255. [See below](#container-networking-and-application-security-groups)
 
-* `require_proxied_app_traffic`: Set this to `true` if Diego was configured to require proxied port mappings, i.e. if `containers.proxy.enable_unproxied_port_mappings` is set to `false`.
-
 * `staticfile_buildpack_name` [See below](#buildpack-names)
 * `java_buildpack_name` [See below](#buildpack-names)
 * `ruby_buildpack_name` [See below](#buildpack-names)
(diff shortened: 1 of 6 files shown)


jochenehret (approved):
Tested on cf-deployment v43.2.0, works. Also tested with PR https://github.com/cloudfoundry/cf-deployment/pull/1195 merged, works!

jochenehret:
I'll wait one more day to give the other ARD WG members a chance for review, then I'll merge the PR.

davewalter (approved):
Looks good.

ctlong:
Hmmmm are we okay with losing the present/not present check on the env vars?

Currently, if `require_proxied_app_traffic` is `true` then we expect `CF_INSTANCE_ADDR` and `CF_INSTANCE_PORT` to be set. These changes result in the test ignoring them if they're not set, regardless of other circumstances.

davewalter:
Fair point.

geofffranks:
We can switch the default value, but then all the pipelines using older cd-d with the latest CATs will require manual intervention. If we don't, all the newer cf-d with any copy of CATs will require manual intervention.

The reasoning for removing these checks was that they're very diego specific, and already tested in unit + integration tests by diego's pipelines. When digging back through the history of when they were added, it seemed like they were added to catch any specific problems that could trip up apps, but instead as satisfying a "this feature needs acceptance tests" checkbox. 

We discussed internally, and we were OK with increased risk by removing these tests as a trade off to making everyone reconfigure CATs to have it continue to work.

ctlong (approved):

PR #1239, opened 2024-09-10
Add acceptance test for crashed app instance stats

kathap:
When pushing an app which has crashing instances (e.g. last instance crashes) all instances should be reported with their correct state.

Add a test when pushing an app which has crashing instances (e.g. last instance crashes) check that the states of the instances are reported correctly.

[#3953](https://github.com/cloudfoundry/cloud_controller_ng/pull/3953)

v42.3.0

- [ ] YES
- [X] N/A

~90 seconds

@johha @jochenehret

### DIFF
--- a/apps/crashing.go
+++ b/apps/crashing.go
@@ -16,6 +16,32 @@ import (
 	"time"
 )
 
+func hasOneInstanceInState(appName, processPath, desiredState string) bool {
+	// Perform the CF curl command to get process stats
+	session := cf.Cf("curl", processPath).Wait()
+
+	// Parse the JSON response
+	instancesJson := struct {
+		Resources []struct {
+			Type  string `json:"type"`
+			State string `json:"state"`
+		} `json:"resources"`
+	}{}
+
+	// Read the session output and unmarshal the JSON data
+	bytes := session.Wait().Out.Contents()
+	err := json.Unmarshal(bytes, &instancesJson)
+	Expect(err).ToNot(HaveOccurred(), "Error unmarshalling process stats JSON")
+
+	// Check if any instance is in the desired state
+	for _, instance := range instancesJson.Resources {
+		if instance.State == desiredState {
+			return true
+		}
+	}
+	return false
+}
+
 var _ = AppsDescribe("Crashing", func() {
 	var appName string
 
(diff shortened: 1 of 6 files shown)


assets/python-crash-app/requirements.txt
```diff
@@ -0,0 +1 @@
+# requirements.txt
```
jochenehret, line 1: Do we really need this empty file?

apps/crashing.go
```diff
@@ -45,6 +53,69 @@ var _ = AppsDescribe("Crashing", func() {
 		})
 	})
 
+	Describe("an app with three instances, two running and one crashing", func() {
+		It("keeps two instances running while another crashes", func() {
+			By("Pushing the app with three instances")
+			Expect(cf.Cf(
+				"push", appName,
```
jochenehret, line 60: Some parameters are already defined in the test app's manifest, do we need both?

kathap: Not really..

apps/crashing.go
```diff
@@ -45,6 +47,69 @@ var _ = AppsDescribe("Crashing", func() {
...
+			By("Checking that the app is up and running")
+			Eventually(cf.Cf("app", appName)).Should(Say("running"))
+
+			By("Waiting until one instance crashes")
+			appGuid := app_helpers.GetAppGuid(appName)
+			processStatsPath := fmt.Sprintf("/v3/apps/%s/processes/web/stats", appGuid)
+
+			// Poll until at least one instance has crashed
+			Eventually(func() bool {
+				session := cf.Cf("curl", processStatsPath).Wait()
```
jochenehret, line 70: Could this block be extracted into a function? Then you could call
```
hasOneInstanceState("CRASHED")
...
hasOneInstanceState("RUNNING")
```

kathap: Good catch

johha (dismissed):

apps/crashing.go
```diff
@@ -1,19 +1,47 @@
...
 	"github.com/cloudfoundry/cf-test-helpers/v2/cf"
 	"github.com/cloudfoundry/cf-test-helpers/v2/helpers"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+	"time"
 )
 
+func hasOneInstanceInState(appName, processPath, desiredState string) bool {
```
jochenehret, line 19: "appName" parameter is not used in function?

jochenehret (changes requested):

jochenehret (approved):

johha (approved):

PR #1257, opened 2024-09-26
Re-enable volume service test

jochenehret:

Fix the volume service acceptance test so that it works with the cf-deployment NFS service.

We want to validate the NFS service feature that is offered by cf-deployment:
https://github.com/cloudfoundry/cf-deployment/blob/main/operations/enable-nfs-volume-service.yml
The cf-deployment "fresh" environment will deploy the NFS broker and a test server:
https://github.com/cloudfoundry/cf-deployment/pull/1204
Then we can activate the volume_service test on that environment:
https://github.com/cloudfoundry/relint-envs/pull/46

v43.3.0

- [x] introduces a new test --- Are you sure everyone should be running this test? -> test must be explicitly enabled

- [x] YES
- [ ] N/A

It's not strictly a new test, we want to re-enable an existing test.

150 seconds


jochenehret:
CATs config json would need these parameters:
```
  "volume_service_name": "nfs",
  "volume_service_plan_name": "Existing",
  "volume_service_create_config": "{\"share\": \"nfstestserver.service.cf.internal/export/users\"}"
```

ctlong (approved):

winkingturtle-vmw:
We've now pulled in these changes, and verified that above changes will succeed for SMB broker as well.

PR #1266, opened 2024-10-07
Add Option for  Volume Service Broker name

winkingturtle-vmw:

Add Option for  Volume Service Broker name

If VolumeService Service broker is configured with a different broker-name, cats will fail. With this change, it'll be required to provide a `volume_service_broker_name` only when running volume-services tests

main

- [ ] YES
- [x] N/A

N/A

N/A

- [x] **Urgent** - unblocks current or future work (our pipeline's are currently broken)

### DIFF
--- a/helpers/config/config.go
+++ b/helpers/config/config.go
@@ -95,6 +95,7 @@ type CatsConfig interface {
 	GetVolumeServicePlanName() string
 	GetVolumeServiceCreateConfig() string
 	GetVolumeServiceBindConfig() string
+	GetVolumeServiceBrokerName() string
 
 	GetReporterConfig() reporterConfig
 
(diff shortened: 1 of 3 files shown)


davewalter (commented):
@winkingturtle-vmw Is the intention for this to be a required configuration when running the volume services tests? If so, it would be good to add a check to the [validateVolumeServices](https://github.com/cloudfoundry/cf-acceptance-tests/blob/55b0b5c05b74d1bdd9f9bd3b7f8d4262ffb6a2b7/helpers/config/config_struct.go#L699-L716) helper function to ensure correct configuration. If not, then I think the changes to `volume_services/volume_services.go` need to be adjusted to not provide the `-b` flag to `enable-service-access` and `create-service` when the service broker name is empty.

davewalter (approved):

davewalter:
> I've updated this PR to have a backward compatible behavior. Currently, we are running this PR through our pipeline to see if we missed anything.

Thanks @winkingturtle-vmw. Please let us know if your CI went green and we can go ahead and merge this.

winkingturtle-vmw:
Above changes has gone green in our pipeline.

Issue #1276, opened 2024-10-14
Fresh CATS fails due to: "curl: (92) HTTP/2 stream 1 was not closed cleanly: PROTOCOL_ERROR (err 1) [HTTP2]"


dimivel:
# Issue
Currently the [fresh-cats job](https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/fresh-cats/builds/1548) is failing because of the Routing Transparency test that should appropriately handles certain reserved/unsafe characters but its not the case.

This is the only foundation that is enabled with http2 as default backend protocol using the GCP classical Application Load Balancer.

# Context
This bbl env is enabled with [http2](https://github.com/cloudfoundry/relint-envs/blob/main/environments/test/luna/bbl-state/terraform/http2-cf-lb-gcp_override.tf) as backend protocol. As of today, the last 5 consecutive build resulted in red jobs because of the unit test above mentioned is failing. We see the following error in the builds:

```
  [FAILED] Expected
      <int>: 92
  to match exit code:
      <int>: 0
```

If manually, tested against this foundation using the [golang app](https://github.com/cloudfoundry/cf-acceptance-tests/tree/main/assets/golang), we get the following error for curl for two reserved/unsafe characters: the `'\'`(backward slash) and `'"'`(double quotes) :
 
```
curl: (92) HTTP/2 stream 1 was not closed cleanly: PROTOCOL_ERROR (err 1)
```

# Expected result
The CATS should pass and jobs should be green.

# Current result
Fail to run the CATS against a fresh foundation where http2 backend protocol is enabled for a a GCP classical Application Load balancer.
 
# Possible Fix
1. Remove the two problematic characters (Now temporary work-around)
2. (Not confirmed) Confirm with routing team, if there gorouter is properly handling the reserved/unsafe characters.
3. (Partially confirmed) Issue with GCP??, open ticket on them, keep the same unit tests.

dimivel:
hey Carson,

If we force the curl to use the http1.1 it works, output
```
curl 'https://cat6-go-test.cf.luna.env.wg-ard.ci.cloudfoundry.org/requesturi/!~^\"()$?!()$#!' --http1.1 -v
*   Trying 34.111.232.161:443...
* Connected to cat6-go-test.cf.luna.env.wg-ard.ci.cloudfoundry.org (34.111.232.161) port 443 (#0)
* ALPN, offering http/1.1
[...]

PR #1277, opened 2024-10-14
Remove two problematic characters from Routing Transparency test

dimivel:

Remove the two problematic characters from Routing Transparency test that should appropriately handles certain reserved/unsafe characters but its not the case.

https://github.com/cloudfoundry/cf-acceptance-tests/issues/1276

v44.0.0

- [ ] YES
- [x] N/A

### DIFF
--- a/apps/routing_transparency.go
+++ b/apps/routing_transparency.go
@@ -44,11 +44,11 @@ var _ = AppsDescribe("Routing Transparency", func() {
 	})
 
 	It("appropriately handles certain reserved/unsafe characters", func() {
-		curlResponse := helpers.CurlApp(Config, appName, "/requesturi/!~^'()$\"?!'()$#!'")
+		curlResponse := helpers.CurlApp(Config, appName, "/requesturi/!~^'()$?!'()$#!'")
 		Expect(curlResponse).To(ContainSubstring("Request"))
 
 		By("preserving all characters")
-		Expect(curlResponse).To(ContainSubstring("/requesturi/!~^'()$\""))
+		Expect(curlResponse).To(ContainSubstring("/requesturi/!~^'()$"))
 		Expect(curlResponse).To(ContainSubstring("Query String is [!'()$]"))
 	})
 })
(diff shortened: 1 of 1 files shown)


apps/routing_transparency.go
```diff
@@ -44,11 +44,11 @@ var _ = AppsDescribe("Routing Transparency", func() {
 	})
 
 	It("appropriately handles certain reserved/unsafe characters", func() {
-		curlResponse := helpers.CurlApp(Config, appName, "/requesturi/!~^'()$\"?!'()$#!'")
```
jochenehret, line 47: Better solution: Check if `GetIncludeHTTP2Routing()` is `false` and only then run the double-quote test.

dimivel: Done, please review if our intention is correct.

jochenehret: That would skip the whole test if HTTP/2 is enabled. What I meant is just skipping the " test and only enable it if HTTP/2 is disabled. Something like this (not tested):
```
	It("appropriately handles certain reserved/unsafe characters", func() {
		curlResponse := helpers.CurlApp(Config, appName, "/requesturi/!~^'()$?!'()$#!'") // no " in request uri
		Expect(curlResponse).To(ContainSubstring("Request"))

		By("preserving all characters")
		Expect(curlResponse).To(ContainSubstring("/requesturi/!~^'()$"))
		Expect(curlResponse).To(ContainSubstring("Query String is [!'()$]"))

		By("preserving double quotes (not for HTTP/2)")
		if !Config.GetIncludeHTTP2Routing() {
			curlResponse = helpers.CurlApp(Config, appName, "/requesturi/\"")
			Expect(curlResponse).To(ContainSubstring("/requesturi/\""))
		}
	})
```

dimivel: The proposal above is tested locally and works as expected.

jochenehret (approved):

ctlong (approved):

PR #1291, opened 2024-10-29
Set the cats-config defaults and missing values

iaftab-alam:

Update the example-cats-config.json with all available options and set the defaults correctly.

Update the example-cats-config.json with all available options and set the defaults correctly.

Some tests have been improved and new configuration options have been introduced. Similarly, not all options are added to the CATs integration-config example file. This PR aim to add the defaults and all the missing values.

N/A

- [ ] YES
- [x] N/A

N/A

N/A


jochenehret (commented):
Ok now the file has become quite big. I didn't realize that we have so many options. Perhaps we should add this file as an additional `complete-cats-config.json`? Any opinions?

iaftab-alam:
yes! It make sense, suggesting the `complete-cats-config.json` file for all CATs options is a great idea! It will provide a comprehensive reference for all CATs configurations in one place without digging through code to find the default values.

complete-cats-config.json
```diff
@@ -0,0 +1,92 @@
+IN DEVELOPMENT
```
jochenehret, line 1: I think we can remove "IN DEVELOPMENT" for both files.

jochenehret (changes requested):
"IN DEVELOPMENT" can be removed after this overhaul. Let's wait a few more days to give everyone a chance for review.

complete-cats-config.json
iaftab-alam: done

davewalter:
> yes! It make sense, suggesting the `complete-cats-config.json` file for all CATs options is a great idea! It will provide a comprehensive reference for all CATs configurations in one place without digging through code to find the default values.

Are we concerned that this file will get out of date as we modify the configuration in the future? I am especially concerned about the default values changing and folks not realising that the sample file is wrong.

iaftab-alam:
> > yes! It make sense, suggesting the `complete-cats-config.json` file for all CATs options is a great idea! It will provide a comprehensive reference for all CATs configurations in one place without digging through code to find the default values.
> 
> Are we concerned that this file will get out of date as we modify the configuration in the future? I am especially concerned about the default values changing and folks not realising that the sample file is wrong.

“Good point! To help avoid drift and make sure the file remains up-to-date, I suggest making it part of the PR process to verify this JSON file and make sure all the values are up-to-date. Adding something a checklist item in PRs template or even a checker on PR to verify if configuration changes is impacting this json file. Lastly, any of the approvers can also check this [manually] and enforce the policy.

davewalter:
> “Good point! To help avoid drift and make sure the file remains up-to-date, I suggest making it part of the PR process to verify this JSON file and make sure all the values are up-to-date. Adding something a checklist item in PRs template or even a checker on PR to verify if configuration changes is impacting this json file. Lastly, any of the approvers can also check this [manually] and enforce the policy.

My preference would be to have a small Golang utility that could generate a complete config.json on demand from the existing defaults in the code. Users could then run it and modify the resulting JSON file as desired for their environment. I feel like adding a manual verification/enforcement step to the PR process is not going to be enough.

PR #1312, opened 2024-11-04
Adapt SSO and ASGs test for disabled v2 setup

jochenehret:

Adapt SSO and ASGs test for a CF deployment with disabled v2 API.

Part of the v2 deprecation story: https://github.com/cloudfoundry/community/pull/941

v44.5.0 with `temporary_enable_v2` set to `false`

- [ ] YES
- [x] N/A

No change in runtime behaviour.

### DIFF
--- a/security_groups/dynamic_asgs.go
+++ b/security_groups/dynamic_asgs.go
@@ -46,7 +46,7 @@ var _ = SecurityGroupsDescribe("ASGs", func() {
 	})
 
 	It("manages whether apps can reach certain IP addresses per ASG configuration", func() {
-		proxyRequestURL := fmt.Sprintf("%s%s.%s/https_proxy/cloud-controller-ng.service.cf.internal:9024/v2/info", Config.Protocol(), appName, Config.GetAppsDomain())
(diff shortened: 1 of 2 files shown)


security_groups/dynamic_asgs.go
```diff
@@ -46,7 +46,7 @@ var _ = SecurityGroupsDescribe("ASGs", func() {
 	})
 
 	It("manages whether apps can reach certain IP addresses per ASG configuration", func() {
-		proxyRequestURL := fmt.Sprintf("%s%s.%s/https_proxy/cloud-controller-ng.service.cf.internal:9024/v2/info", Config.Protocol(), appName, Config.GetAppsDomain())
+		proxyRequestURL := fmt.Sprintf("%s%s.%s/https_proxy/cloud-controller-ng.service.cf.internal:9024/", Config.Protocol(), appName, Config.GetAppsDomain())
```
ctlong, line 49: Is this an info endpoint on v2 and v3? Or just on v3?

jochenehret: The root endpoint is always available, independent of the `temporary_enable_v2` setting. The test just checks with a simple regex match if the endpoint can be reached. The "cloud_controller_v3" key is always present: https://github.com/cloudfoundry/cloud_controller_ng/blob/4c3458dcaf4c1f8fef09a3c316905b9900b4e025/app/controllers/runtime/root_controller.rb#L18

ctlong (approved):

PR #1371, opened 2024-12-09
CI: Add GH actions workflow

ctlong:

Yes

Add a GH actions workflow for CI purposes.

None

N/A

- [ ] YES
- [x] N/A

N/A

None

None

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,8 +5,6 @@ on:
     branches: [ "develop", "release-candidate", "main" ]
   pull_request:
     branches: [ "develop", "release-candidate", "main" ]
-  schedule:
-  - cron: '00 7 * * 1'
 
 jobs:
   unit-test:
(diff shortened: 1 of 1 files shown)


ctlong:
Part of #1372

davewalter (approved):
Looks good. The only question I have is whether this workflow needs to run on a schedule in addition to when changes are made to the core branches?

ctlong:
> Looks good. The only question I have is whether this workflow needs to run on a schedule in addition to when changes are made to the core branches?

It doesn't need to run on a schedule. I added the schedule out of habit, as many of the GH actions workflows I've written in the past contain jobs that may change over time (e.g. linting). These specific jobs should be consistent without code changes though, so I'll remove the schedule.

davewalter (approved):

ctlong:
The revert commit above was accidental. I've deleted the branch now.

Issue #1352, opened 2024-11-28
RFC0027: Add acceptance test for per-route features


maxmoehl:
Since we have implemented most of the RFC by now and the CC implementation is getting close to being done we also want to add an acceptance test. This issue tracks progress on implementing the tests.

maxmoehl:
PR got merged!

PR #1424, opened 2025-01-20
test: add test for loadbalancing route option

maxmoehl:

Add an acceptance test for per-route options as described in #1352.

#1352, cloudfoundry/community#909

-

- [x] YES
- [ ] N/A

It tests the E2E functionality of per-route options. Each component has its own acceptance tests but we want to ensure that the E2E scenario works with every released cf-d.

Unknown


routing/per_route_options.go
```diff
@@ -0,0 +1,129 @@
...
+				"--var", fmt.Sprintf("domain=%s", Config.GetAppsDomain()),
+				"--var", fmt.Sprintf("name=%s", appName),
+				"--var", fmt.Sprintf("leastconnhost=%s", leastConnHost),
+				"--var", fmt.Sprintf("roundrobinhost=%s", roundRobinHost),
+				"-f", filepath.Join(asset.Dora, "man_route_opt_lb.yml"),
+			).Wait(Config.CfPushTimeoutDuration())).To(Exit(0))
+			appId = app_helpers.GetAppGuid(appName)
+			appInstanceRegex, _ := regexp.Compile("[[:alnum:]]{8}(-[[:alnum:]]{4}){4}")
+			for i := range 2 {
+				for {
```
ctlong, line 53: Can you please change this infinite for loop to be an `Eventually` block instead. That comes with two benefits:
- Times out eventually, if it's not succeeding
- Obviates the need for a manual `time.Sleep`

routing/per_route_options.go
```diff
@@ -0,0 +1,129 @@
...
+		Context("when it's set to round-robin", func() {
+			It("distributes requests evenly", func() {
+				doraUrl := fmt.Sprintf("%s%s.%s", Config.Protocol(), roundRobinHost, Config.GetAppsDomain())
+				var wg sync.WaitGroup
+				for i := 0; i < 10; i++ {
+					wg.Add(1)
+					go func() {
+						defer wg.Done()
+						defer GinkgoRecover()
+						Expect(helpers.Curl(Config, fmt.Sprintf("%s/delay/20", doraUrl), "-H", fmt.Sprintf("X-Cf-App-Instance: %s:0", appId)).Wait(25 * time.Second)).To(Exit(0))
```
ctlong, line 82: This expectation will likely behave strangely due to the asynchronous nature. Is there some other mechanism you can use here instead?

routing/per_route_options.go
```diff
@@ -0,0 +1,129 @@
...
+					}()
+					time.Sleep(10 * time.Millisecond)
```
ctlong, line 84: Why is this sleep necessary?

routing/per_route_options.go
```diff
@@ -0,0 +1,129 @@
...
+				}
+
+				reqCount := [2]int{0, 0}
+				for i := 0; i < 20; i++ {
+					id := helpers.Curl(Config, fmt.Sprintf("%s/id", doraUrl)).Wait().Out.Contents()
+					reqCount[slices.Index(instanceIds[:], string(id))] += 1
+					time.Sleep(10 * time.Millisecond)
+				}
```
ctlong, line 92: This whole for loop strikes me as strange. Can you please elaborate what the purpose is? If possible, I'd like to remove the `time.Sleep`. Could the `WaitGroup` help here?

ctlong (changes requested):

maxmoehl:
Those things are mostly leftovers from our testing to figure out what was wrong. I'll clean them up.

routing/per_route_options.go
maxmoehl: We need to find out whether requests are load-balanced they way we expect. To do so we request the apps instance ID and increment the number of requests served by that instance. I've removed the sleep, those were mostly because we were initially suspecting a overload on the app but that doesn't seem to be the case.

routing/per_route_options.go
maxmoehl: We could drop the expect and ignore the return value from the cURL, e.g.:

```suggestion
						helpers.Curl(Config, fmt.Sprintf("%s/delay/20", doraUrl), "-H", fmt.Sprintf("X-Cf-App-Instance: %s:0", appId))
```

Not sure what else we could do here as we need those requests to stay active and cause the imbalance between instances needed to check the algorithm used by gorouter.

maxmoehl:
Once we are happy with the code we can have another test run to ensure everything still works properly. @ctlong please have another look.

assets/dora/man_route_opt_lb.yml
```diff
@@ -0,0 +1,24 @@
+---
+applications:
+- name: ((name))
```
ctlong, line 3: You don't need the `name` var. The `cf push APP_NAME...` command should overwrite whatever the app name is set to in the manifest.

Please replace this with a hardcoded name, e.g. "dora"

assets/dora/man_route_opt_lb.yml
```diff
@@ -0,0 +1,24 @@
+---
```
ctlong, line 1: Can you please rename the manifest to include the word "manifest". There are many other manifests in CATs and typically they all have "manifest" in their filename to make it easy to search for them all.

How's `route_options_manifest.yml`?

routing/per_route_options.go
```diff
@@ -0,0 +1,125 @@
...
+				"-m", DEFAULT_MEMORY_LIMIT,
+				"-p", asset.Dora,
+				"--var", fmt.Sprintf("domain=%s", Config.GetAppsDomain()),
+				"--var", fmt.Sprintf("name=%s", appName),
+				"--var", fmt.Sprintf("leastconnhost=%s", leastConnHost),
+				"--var", fmt.Sprintf("roundrobinhost=%s", roundRobinHost),
+				"-f", filepath.Join(asset.Dora, "man_route_opt_lb.yml"),
+			).Wait(Config.CfPushTimeoutDuration())).To(Exit(0))
+			appId = app_helpers.GetAppGuid(appName)
+			appInstanceRegex, _ := regexp.Compile("[[:alnum:]]{8}(-[[:alnum:]]{4}){4}")
```
ctlong, line 50: Should use `regexp.MustCompile` here instead.

ctlong, line 50: Should use `regexp.MustCompile` here instead.

routing/per_route_options.go
```diff
@@ -0,0 +1,125 @@
...
+				var wg sync.WaitGroup
+				for i := 0; i < 10; i++ {
+					wg.Add(1)
+					go func() {
+						defer wg.Done()
+						defer GinkgoRecover()
+						helpers.Curl(Config, fmt.Sprintf("%s/delay/20", doraUrl), "-H", fmt.Sprintf("X-Cf-App-Instance: %s:0", appId))
+					}()
+				}
+
```
ctlong, line 85: I think `wg.Wait()` should be here. That way you don't go trying to count the number of times the instances were curled until after the curls have completed.

routing/per_route_options.go
```diff
@@ -0,0 +1,125 @@
...
+		Context("when it's set to round-robin", func() {
+			It("distributes requests evenly", func() {
+				doraUrl := fmt.Sprintf("%s%s.%s", Config.Protocol(), roundRobinHost, Config.GetAppsDomain())
+				var wg sync.WaitGroup
+				for i := 0; i < 10; i++ {
+					wg.Add(1)
+					go func() {
+						defer wg.Done()
+						defer GinkgoRecover()
+						helpers.Curl(Config, fmt.Sprintf("%s/delay/20", doraUrl), "-H", fmt.Sprintf("X-Cf-App-Instance: %s:0", appId))
```
ctlong, line 82: Wouldn't using the `X-Cf-App-Instance` header here necessitate that every request goes to that specific instance. Why not remove it?

ctlong: Also, for round-robin you don't need to use the `delay/20` endpoint right? Removing that will speed up this test run.

ctlong (changes requested):

routing/per_route_options.go
ctlong: In fact, why do you need this request spamming at all in this case. In order to test that round-robin is occurring you just need the following for loop to result in roughly each requests to each process, right?

maxmoehl: We are creating these requests because if there is no traffic to dora round-robin and least-connection will behave identical. So we enforce a situation where one instance has ten ongoing requests while the other has zero. That way we can observe the difference between round-robin and least-connection.

Re: wait group: If we did the wait group in line 85 we would only test the algorithm once those requests have finished and both instances are back to zero connections. I'll tighten the timings as the test should run fast but I don't really see a way to avoid it completely.

routing/per_route_options.go
maxmoehl: See https://github.com/cloudfoundry/cf-acceptance-tests/pull/1424#discussion_r1974873228

ctlong (dismissed):

maxmoehl:
Thanks! @iaftab-alam ran the tests again and they pass so this is good to merge from our side.

hoffmaen:
> LGTM!
> 
> Can you please record how many minutes this will increase CATs runs? You can do this by comparing the previous runtime of the routing test suite with the runtime of the new routing test suite, using around 12 parallel nodes.
> 
> And can you also list the CF-D version that you ran this against? Actually... does this require some minimum CF-D/CAPI version?

Hi @ctlong , we executed CATs on both this branch, and the `main` branch three times each with the following setup:
* `cf-deployment v48.4.0` (`capi v1.203.0`)
* 12 nodes

The per-route load-balancing options tests require `cf-deployment >= v48.1.0`.

Results:
* develop branch (with load-balancing test):
** 1st run: 15m35s 
** 2nd run: 17m22s
** 3rd run: 15m41s
--> Average 16m21s, Median 15m41s

* `main` branch:
** 1st run: 14m33s
** 2nd run: 15m44s
** 3rd run: 15m25s
--> Average 15m23s, Median 15m25s

So the additional tests will take approximately 1 minute, when comparing the average and/or fastest execution time (and ignoring the outlier with >17m). As we deploy DORA twice, this is expected.

ctlong (approved):

PR #1472, opened 2025-02-27
ci(build-catnip-docker-image): Apply latest version over a commit SHA

ctlong:

Yes

Change a GH Actions Workflow (build-catnip-docker-image) to version docker/build-push-action instead of using a commit SHA.

None

N/A

- [ ] YES
- [x] N/A

N/A

None

None

### DIFF
--- a/.github/workflows/build-catnip-docker-image.yml
+++ b/.github/workflows/build-catnip-docker-image.yml
@@ -28,7 +28,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Build and push Docker image
         id: push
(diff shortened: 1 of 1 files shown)


ctlong:
As far as I can tell, we should version `docker/build-push-action` because it is released fairly regularly and the latest release covers the commit SHA previously specified.

Unless there's some reason we want to stay pinned to this SHA?

ctlong:
I didn't attempt to version the other `docker` action because it's not released as frequently, so I can see some benefit in just using the latest commit SHA.

davewalter (approved):
I assume that Dependabot will bump this when new versions are released?

PR #1474, opened 2025-03-01
fix: Bump the minimum required version of Go to go1.23.0

ctlong:

Yes

Updating the minimum required version of Go to go1.23.0.

If any dependencies of a module require a minimum version of Go that is greater than the minimum version of Go required by the module, then `go mod tidy` will update the minimum version of the module. This is what was happening when I ran `go mod tidy`. I've simply committed the result of that command.

v48.3.0

- [x] YES
- [ ] N/A

N/A

None

None

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,8 @@
 module github.com/cloudfoundry/cf-acceptance-tests
 
-go 1.22
-toolchain go1.23.6
+go 1.23.0
+
+toolchain go1.24.0
 
 require (
 	code.cloudfoundry.org/archiver v0.26.0
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ in such a way as to impact other tests.
 
 ## Test Setup
 ### Prerequisites for running CATS
-- Install golang >= `1.22`. Set up your golang development environment, per
+- Install golang >= `1.23.0`. Set up your golang development environment, per
   [golang.org](http://golang.org/doc/install).
 - Install the [`cf CLI`](https://github.com/cloudfoundry/cli) >= `8.5.0`. Make
   sure that it is accessible in your `$PATH`.
@@ -53,7 +53,7 @@ in such a way as to impact other tests.
 All `go` dependencies required by CATs
 are vendored in the `vendor` directory.
 
-Make sure to have Golang 1.22+
+Make sure to have Golang 1.23+
 
 In order to update a current dependency to a specific version,
 do the following:


dimivel (approved):

oliver-heinrich (approved):
LGTM

go.mod
```diff
@@ -1,7 +1,8 @@
 module github.com/cloudfoundry/cf-acceptance-tests
 
-go 1.22
-toolchain go1.23.6
+go 1.23.0
```
davewalter, line 3: I noticed that the unit tests specifically used Golang 1.23.0 instead of the latest patch release. Does this need to be "go 1.23" to allow CATs to be run with the latest patch release?

ctlong: It seems like the Go community is leaning toward specifying patch versions in their `go` directives, so I'm inclined to do the same. If we do care about making sure that our ci workflow is running unit tests with more recent patches of go1.23+, then we could instead switch our workflows to use [`stable` and/or `oldstable`](https://github.com/actions/setup-go?tab=readme-ov-file#using-stableoldstable-aliases) as the Go versions. What do you think?

davewalter: My biggest concern is the version used when running CATs against a deployment. The comment from the [commit that bumped to Golang 1.22](https://github.com/cloudfoundry/cf-acceptance-tests/commit/a0e5ea36d59c6d7ba56767391ab05bcabe2ca81b) seems to imply that there was a reason for us to omit the patch version, although that should mainly affect the Golang assets, rather than the tests themselves.

ctlong: Adding a `go` directive with a patch version to the Go assets broke CATs [when run with a certain offline Go buildpack](https://github.com/cloudfoundry/cf-acceptance-tests/commit/043a63ff66b5a67c36a14876edf9095d74b62783). I wonder if that edge case has now been fixed 🤔...  Anyways, if we can avoid adding the patch version to our Go assets' `go` directive then I guess we should. However, the `go.mod` at the repo root is unaffected by that particular issue, and I've since changed my mind about the reasoning that led me to make them match last time.

I think we probably should enforce that Go assets need to have a minimum Go version that is less than or equal to the minimum required Go version in the `go.mod` at the repo root. Do you think that's best enforced by making them all match?

davewalter: > However, the go.mod at the repo root is unaffected by that particular issue, and I've since changed my mind about the reasoning that led me to make them match last time.

Sounds reasonable.

> I think we probably should enforce that Go assets need to have a minimum Go version that is less than or equal to the minimum required Go version in the go.mod at the repo root.

Bumping the major/minor version of all go.mod files at the same time seems reasonable, since the driving force (support for the old version of Golang) is presumably the same. We just need to ensure that the go.mod files for the test assets do not specify a patch version, where the old for the test suite itself does.

davewalter (approved):

PR #1476, opened 2025-03-05
Add canary instance step test

Samze:

Adds test for Canary steps.

* CAPI change: https://github.com/cloudfoundry/cloud_controller_ng/pull/4219
* CF CLI change: https://github.com/cloudfoundry/cli/pull/3390

* [v48.5.0](https://github.com/cloudfoundry/cf-deployment/releases/tag/v48.5.0)

- [x] YES
- [ ] N/A

Integration of CF CLI and CAPI.

~2mins 

### DIFF
--- a/v3/deployment.go
+++ b/v3/deployment.go
@@ -1,7 +1,7 @@
 package v3
 
 import (
-	"strings"
+	"fmt"
 	"time"
 
 	"github.com/cloudfoundry/cf-acceptance-tests/helpers/app_helpers"
@@ -26,6 +26,8 @@ var _ = V3Describe("deployment", func() {
 		appCheckerIsDone     <-chan bool
 	)
 
+	const numberOfAppCurlChecks = 10
+
 	BeforeEach(func() {
 		if !Config.GetIncludeDeployments() {
 			Skip(skip_messages.SkipDeploymentsMessage)
@@ -37,7 +39,7 @@ var _ = V3Describe("deployment", func() {
 		Eventually(func(g Gomega) {
 			session := cf.Cf("app", appName).Wait()
 			g.Expect(session).Should(Say(`instances:\s+3/3`))
-		})
+		}).Should(Succeed())
 		Eventually(func() string {
 			return helpers.CurlAppRoot(Config, appName)
 		}).Should(ContainSubstring("Hi, I'm Dora"))
@@ -76,19 +78,9 @@ var _ = V3Describe("deployment", func() {
 				g.Expect(session).Should(Exit(0))
 			}).Should(Succeed())
 
-			counter := 0
-			Eventually(func() int {
-				if strings.Contains(helpers.CurlAppRoot(Config, appName), "Hello from a staticfile") {
-					counter++
-				} else {
-					counter = 0
-				}
-				return counter
-			}).Should(Equal(10))
-
-			Consistently(func() string {
-				return helpers.CurlAppRoot(Config, appName)
-			}).ShouldNot(ContainSubstring("Hi, I'm Dora"))
+			for range numberOfAppCurlChecks {
+				Expect(helpers.CurlAppRoot(Config, appName)).To(ContainSubstring("Hello from a staticfile"))
+			}
 		})
 
 		It("can be cancelled and rolls back to the previous app", func() {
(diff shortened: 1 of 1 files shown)


a-b (approved):
LGTM

Samze:
This is now included in https://github.com/cloudfoundry/cf-deployment/releases/tag/v48.5.0 and ready for review

v3/deployment.go
```diff
@@ -330,6 +331,58 @@ var _ = V3Describe("deployment", func() {
...
+				By("Verifying the continue succeeded and we rolled out the new process")
+				counter := 0
+				Eventually(func() int {
+					if strings.Contains(helpers.CurlAppRoot(Config, appName), "Hello from a staticfile") {
+						counter++
+					} else {
+						counter = 0
+					}
+					return counter
+				}).Should(Equal(10))
```
ctlong, line 379: Why not combine this test and the one below it into one `Consistently` that just checks for the app root to return "Hello from a staticfile" every time?

v3/deployment.go
```diff
@@ -330,6 +331,58 @@ var _ = V3Describe("deployment", func() {
...
+		Context("instance-steps", func() {
+			BeforeEach(func() {
+				Expect(cf.Cf("scale", appName, "-i", "4").Wait(Config.CfPushTimeoutDuration())).To(Exit(0))
+			})
+
+			It("deploys an app, transitions to pause and can be continued multiple times and then deploys successfully", func() {
+				By("Pushing a canary deployment")
+				Expect(cf.Cf("push", appName, "--strategy", "canary", "--instance-steps=25,50,75", "--no-wait", "-b", Config.GetStaticFileBuildpackName(), "-p", assets.NewAssets().Staticfile).Wait(Config.CfPushTimeoutDuration())).To(Exit(0))
+
+				for i := 1; i <= 3; i++ {
```
ctlong, line 344: Do you want to include any validation that the step percentage is working? I notice that in the test above for instance, the `cf app` command output is used to gauge how many instances are running vs. starting. Is there some similar output that could be used here?

Samze: Good idea, that output was not actually present when this test was written but has since been added. I'll update.

v3/deployment.go
Samze: This pattern was copied from elsewhere, I think the original thinking was that the problem with consistently is that it is time based and that the app could take a long time to return. Really what we want is to check a certain number of times to give us confidence that this is the *only* thing the app returns. 

This code doesn't really do that since it uses eventually for the affirmative and uses consistently for the negative.

Thinking will change it simply to be number of calls based, something like
```
				for range 10 {
					Expect(helpers.CurlAppRoot(Config, appName)).To(ContainSubstring("Hello from a staticfile"))
				}
```

v3/deployment.go
```diff
@@ -37,7 +39,7 @@ var _ = V3Describe("deployment", func() {
 		Eventually(func(g Gomega) {
 			session := cf.Cf("app", appName).Wait()
 			g.Expect(session).Should(Say(`instances:\s+3/3`))
-		})
+		}).Should(Succeed())
```
Samze, line 42: **note** I also fixed up a bunch of these incorrect eventually statements that wernt functioning.

v3/deployment.go
```diff
@@ -134,42 +116,20 @@ var _ = V3Describe("deployment", func() {
...
-					g.Expect(session).Should(Say(`#1\s+running`))
-					// new deployment
-					g.Expect(session).Should(Say(`#0\s+starting`))
-					g.Expect(session).Should(Say(`#1\s+starting`))
-					g.Expect(session).ShouldNot(Say(`#2\s+starting`))
 					g.Expect(session).Should(Say("Active deployment with status DEPLOYING"))
 					g.Expect(session).Should(Say(`strategy:\s+rolling`))
 					g.Expect(session).Should(Say(`max-in-flight:\s+2`))
-				})
+				}).Should((Succeed()))
```
ctlong, line 122: It might technically work, but your succeed has an extra set of parens around it. Can you please remove them and check the other succeeds as well?

Thanks for making this change!

v3/deployment.go
```diff
@@ -76,19 +78,9 @@ var _ = V3Describe("deployment", func() {
...
-				}
-				return counter
-			}).Should(Equal(10))
-
-			Consistently(func() string {
-				return helpers.CurlAppRoot(Config, appName)
-			}).ShouldNot(ContainSubstring("Hi, I'm Dora"))
+			for range numberOfAppCurlChecks {
+				Expect(helpers.CurlAppRoot(Config, appName)).To(ContainSubstring("Hello from a staticfile"))
+			}
```
ctlong, line 83: If the intent is to check that it succeeds some number of times rather than over some time period, can we parallelize these requests to potentially speed up the tests?

ctlong (changes requested):

v3/deployment.go
Samze: Good idea, will make them run in parallel

Samze: Added

ctlong (approved):

PR #1500, opened 2025-03-21
remove unnneeded global rand.Seed logic

ameowlia:
since go 1.20 rand has been automatically seeded

Yes

Calling rand.Seed globally is no longer needed. Since go 1.20 rand has been automatically seeded. go 1.20 was released in Feb 2022 and is no longer supported.

* https://pkg.go.dev/math/rand@go1.20#Seed

- [ ] YES
- [x] N/A

n/a

micoseconds faster!

### DIFF
--- a/assets/pora/server.go
+++ b/assets/pora/server.go
@@ -297,13 +297,7 @@ func env(res http.ResponseWriter, req *http.Request) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


ctlong (approved):
Happy to take the microseconds back! Thanks!

PR #1501, opened 2025-03-21
update default suite timeout to 24h, as ginkgov2 changed it to 1h

geofffranks:

Yes i did it right this time!

Ginkgo v2 changed the default suite timeout to 1hr which is too short for cats.

We recommend people set it to 24h in the readme

cats ref 562033f1  + cf-d v44.9.0

- [ ] YES
- [x] N/A

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -18,5 +18,5 @@ export RUN_ID
 
 pushd "${CATS_ROOT}" > /dev/null
   echo "Using $(go run github.com/onsi/ginkgo/v2/ginkgo version)"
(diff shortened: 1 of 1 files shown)


ctlong:
Worth noting that this is not a breaking change. I tested the ginkgo CLI and found that it will accept multiple `--timeout` flags, with the last such flag taking precedence. That means if specific scripts are using `./bin/test` with a `--timeout` flag set, their value will take precedence over this setting of `24h`. Effectively, this just changes the default from 1h to 24h.

ctlong (approved):

PR #1509, opened 2025-03-31
 Update file based service bindings according to updated RFC-0030

serdarozerr:
Implemented Test Cases for file based service bindings. We implemented  for `file-based-vcap-services`  option. 

Please check the following link: 
https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0030-add-support-for-file-based-service-binding.md

v48.5.0

Test is active only when `include_file_based_service_bindings` flag is set to true

- [x] YES
- [ ] N/A

We want to test both mode that described in RFC-0030

11m23s


helpers/services/service_helper.go
```diff
@@ -67,3 +68,46 @@ func ValidateFileBasedServicebinding(appName, serviceName, appGuid, serviceGuid
 	checkFileContent("type", "user-provided")
 	checkFileContent("username", "admin")
 }
+
+func ValidateFileBasedVcapServices(appName, serviceName, appGuid, serviceGuid string) {
+	getEncodedFilepath := func() string {
+		path := fmt.Sprintf("/etc/cf-service-bindings/vcap_services")
```
jochenehret, line 74: "path" can be inlined into the next statement

README.md
```diff
@@ -361,7 +361,7 @@ Test Group Name| Description
 `cnb` | Tests our ability to use cloud native buildpacks.
 `detect` | Tests the ability of the platform to detect the correct buildpack for compiling an application if no buildpack is explicitly specified.
 `docker`| Tests our ability to run docker containers on Diego and that we handle docker metadata correctly.
-`file-based service bindings`| Tests file-based service bindings for a buildpack app, a CNB app and a Docker app.
+`file-based service bindings`| Tests file-based service bindings for a buildpack app, a CNB app and a Docker app. This test group being run against 2 different feature flags on 2 different stacks(windows,cflinuxfs4). For more detail about the feauture flags check [RFC0030](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0030-add-support-for-file-based-service-binding.md)
```
jochenehret, line 364: We shouldn't hard-code the "cflinuxfs4" stack name here. Just mention that the test runs on Linux and Windows.

file_based_service_bindings/life_cycles.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/cloudfoundry/cf-test-helpers/v2/cf"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gexec"
+)
+
+type LifeCycle interface {
+	Prepare(string, string, string) (string, string)
+}
+
+type BuildpackLifecycles struct{}
```
jochenehret, line 18: Better use singular here?
`BuildpackLifecycle struct{}`

jochenehret (changes requested):

jochenehret (dismissed):

ctlong (commented):
Can you please change the `windows/file_based_service_bindings.go` file to use a `JustBeforeEach` / `BeforeEach` for setup instead of this `testPrepare` function.

https://onsi.github.io/ginkgo/#separating-creation-and-configuration-justbeforeeach

PR #1535, opened 2025-04-23
Add IPv6 validation within python and node apps

Milena-Encheva:

Yes.

 The change is related to IPv6 support validation.  The main purpose of the newly created `ipv6` test group is to test IPv6 egress calls using different buildpacks, including Java, Python, and others relevant to our ecosystem which will be added later on.
 
```
v48.9.0
```

- [x] requires an update to a CATs integration-config (must set new flag if you want to run the test)

- [x] YES
- [ ] N/A

_CATs should validate IPv6 egress calls after support of the protocol is provided.

Around 90 seconds per test.

@oliver-heinrich


helpers/config/config_struct.go
```diff
@@ -101,6 +101,7 @@ type config struct {
 	IncludeV3                       *bool `json:"include_v3"`
 	IncludeVolumeServices           *bool `json:"include_volume_services"`
 	IncludeZipkin                   *bool `json:"include_zipkin"`
+	IncludeIpv6                     *bool `json:"include_ipv6"`
```
peanball, line 104: I would prefer `IncludeIPv6` (see also other acronyms such as `IncludeHTTP2Routing`m `IncludeTCPRouting`, etc).

assets/python/server.py
```diff
@@ -1,23 +1,132 @@
...
 from socketserver import ThreadingMixIn
-import os
+
+ENDPOINT_TYPE_MAP = {
+            'api.ipify.org': "IPv4",
+            'api6.ipify.org': "IPv6",
+            'api64.ipify.org': "Dual stack"
+        }
+
+PORT = '8080'
```
peanball, line 16: maybe this could be called `DEFAULT_PORT`, as it is the fallback when there is no `PORT` env variable (e.g. when debugging locally)

assets/python/server.py
```diff
@@ -1,23 +1,132 @@
...
+            message =  "Hello python, world!"
+            self.wfile.write(message.encode('utf-8'))
+            self.wfile.write('\n'.encode('utf-8'))
+        
+    def handle_ipv6_test(self):
         self.send_response(200)
         self.end_headers()
-        message =  "Hello python, world!"
+        tester = IPv6Tester(list(ENDPOINT_TYPE_MAP.keys()))
+        tester.test_all_addresses()
```
peanball, line 115: Here you could have a boolean and output information, and potentially even the status code (e.g. not 200) when the conditions of the test fail.

Additionally, the output of what was provided back from the egress calls, i.e. IPv4, IPv6 addresses, or what failed would be really nice for debugging when the test actually fails.

helpers/config/config_test.go
```diff
@@ -88,6 +88,7 @@ type testConfig struct {
 	IncludeVolumeServices           *bool `json:"include_volume_services,omitempty"`
 	IncludeZipkin                   *bool `json:"include_zipkin,omitempty"`
 	IncludeWindows                  *bool `json:"include_windows,omitempty"`
+	IncludeIpv6         			*bool `json:"include_ipv6,omitempty"`
```
peanball, line 91: The indentation is off because tabs are used and above spaces are used. This can be avoided by using `go fmt`.

helpers/skip_messages/skip_messages.go
```diff
@@ -12,6 +12,7 @@ NOTE: Ensure CNB lifecycle is enabled on your platform before enabling this test
 const SkipFileBasedServiceBindingsBuildpackApp = `Skipping this test because config.IncludeFileBasedServiceBindings is set to 'false'.`
 const SkipFileBasedServiceBindingsCnbApp = `Skipping this test because config.IncludeFileBasedServiceBindings and/or config.IncludeCNB are set to 'false'.`
 const SkipFileBasedServiceBindingsDockerApp = `Skipping this test because config.IncludeFileBasedServiceBindings and/or config.IncludeDocker are set to 'false'.`
+const SkipIncludeIpv6 = `Skipping this test because config.IncludeIpv6 is set to 'false'.`
```
peanball, line 15: I suggest calling this `SkipIPv6`, which makes it clearer what is actually skipped.

ipv6/ipv6.go
```diff
@@ -0,0 +1,57 @@
...
+                        return helpers.CurlApp(Config, appName, "/")
+                    }).Should(ContainSubstring("Hello"))
+
+                    /* This test is checking IPv6 egress calls.
+                    It examines that after making a request to a predifined route,
+                    IPv6 tests are executed successfully */
+                    
+                    Eventually(func() string {
+                        return helpers.CurlApp(Config, appName, "/ipv6-test")
+                    }).Should(ContainSubstring("IPv6 tests executed"))
```
peanball, line 52: There is output for `IPv4 tests succeeded`, `IPv6 tests succeeded`, `Dual stack tests succeeded`. It would be good to have all of those as part of the response and not (just) the stdout of the test app. 

They could then be explicit assertions in the test. This makes it easier to see what exactly failed and why, when something failed.

peanball (changes requested):
The biggest request is to not make the app crash (`sys.exit(1)`) when the IPv6 test fails.

A more fine-grained response from the test endpoint with all the details (IPv4 tests success or failed, what IP was seen, same for v6 and) would be really helpful to see what failed when the test fails. Otherwise you just see "the test failed" without any indication why.

Please run `go fmt` on all Go files.

ipv6/ipv6.go
```diff
@@ -0,0 +1,86 @@
...
+	BeforeEach(func() {
+		appName = random_name.CATSRandomName("APP")
+	})
+
+	AfterEach(func() {
+		app_helpers.AppReport(appName)
+		Expect(cf.Cf("delete", appName, "-f", "-r").Wait()).To(Exit(0))
+	})
+
+	ENDPOINT_TYPE_MAP := map[string]struct {
```
jochenehret, line 28: Should be camelcase: `EndpointTypeMap`

ipv6/ipv6.go
```diff
@@ -0,0 +1,86 @@
...
+				Expect(response).To(ContainSubstring("Hello"))
+			} else {
+				Expect(response).To(ContainSubstring(fmt.Sprintf("%s validation resulted in success", data.validationName)))
+			}
+		}
+	}
+
+	Describe("Egress Capability in Apps", func() {
+		for _, stack := range Config.GetStacks() {
+			stack := stack
```
jochenehret, line 71: That's not needed.

jochenehret (approved):

oliver-heinrich (approved):

oliver-heinrich:
[like]  Heinrich, Oliver reacted to your message:
________________________________

dimivel (commented):
LGTM

PR #1553, opened 2025-05-08
Cleanup shared TCP domain after in AfterEach()

jochenehret:

`cf create-shared-domain` is not idempotent and will fail if the domain already exists (e.g. from a previous CATs run that failed for some other reason), so delete the domain in `AfterEach()`.

This CATs run was green and created the domain:
https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/fips-cats/builds/903

The two subsequent CATs runs were red because the domain already existed:
https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/fips-cats/builds/904
https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/fips-cats/builds/905

v49.0.0

- [ ] YES
- [x] N/A

A few seconds to execute `cf delete-shared-domain`.

### DIFF
--- a/tcp_routing/tcp_routing.go
+++ b/tcp_routing/tcp_routing.go
@@ -36,6 +36,12 @@ var _ = TCPRoutingDescribe("TCP Routing", func() {
 		})
 	})
 
+	AfterEach(func() {
+		workflowhelpers.AsUser(TestSetup.AdminUserContext(), Config.DefaultTimeoutDuration(), func() {
+			Expect(cf.Cf("delete-shared-domain", domainName).Wait()).To(Exit())
+		})
+	})
+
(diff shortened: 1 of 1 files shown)


dimivel (approved):

jochenehret:
No fix necessary, the `create-shared-domain` functions are expected to fail. The setup function checks that the call exits, but it doesn't have to exit with status code 0:
https://github.com/cloudfoundry/cf-acceptance-tests/blob/cc7b2ce098a380008359b62a3d628681e1c19b54/tcp_routing/tcp_routing.go#L35

PR #1564, opened 2025-05-13
Ipv6 validation for Golang

Milena-Encheva:

Yes.

The change is part of the ipv6 egress validation cycle. We progressively extend all of the exciting buildpacks relevant to our ecosystem.  In this PR we provide additional endpoints in the Golang application for testing IPv6 support. In case ipv6 validation is enabled, we verify that egress ipv6 call are possible with Go applications. 

v48.9.0

- [ ] YES
- [x] N/A

_CATs should validate IPv6 egress calls with Golang application. We add changes regarding to this buildpack only. The test group for ipv6 was already created.

Around 90 seconds per test.

@oliver-heinrich @iaftab-alam

### DIFF
--- a/assets/golang/site.go
+++ b/assets/golang/site.go
@@ -2,29 +2,119 @@ package main
 
 import (
 	"fmt"
+	"io"
+	"log"
+	"net"
 	"net/http"
 	"os"
+	"time"
 )
 
+type EndpointType struct {
+	validationName string
+	path           string
+}
+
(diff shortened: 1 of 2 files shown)


oliver-heinrich:
Hi @jochenehret , seems that only approvers can add reviewers to the PR. Aftab and Dimitar are currently not available...can you add the reviewers?

dimivel (approved):

PR #1565, opened 2025-05-13
Ipv6 validation for Javaspring apps

Milena-Encheva:

Yes.

The change is part of the ipv6 egress validation cycle. We progressively extend all of the exciting buildpacks relevant to our ecosystem. In this PR we provide additional endpoints in the Java Spring application for testing IPv6 support. In case ipv6 validation is enabled, we verify that egress ipv6 call are possible with Java Spring applications.

v48.9.0

- [ ] YES
- [x] N/A

_CATs should validate IPv6 egress calls with JavaSpring application. We add changes regarding to this buildpack only. The test group for ipv6 was already created.

Around 90 seconds per test.

@oliver-heinrich @iaftab-alam


assets/java-spring/src/main/java/org/cloudfoundry/SpringBootTrivialApplication.java
```diff
@@ -1,19 +1,110 @@
...
+            } else if (inetAddress instanceof java.net.Inet6Address) {
+                return "IPv6";
+            }
+        } catch (Exception e) {
+            logger.severe("Invalid IP format or unexpected error for: " + ipString + ". Error: " + e.getMessage());
+        }
+        return "Invalid IP";
+    }
+
+    private static class EndpointInfo {
```
jochenehret, line 101: This could be simplified with:
```
private static record EndpointInfo(String validationName, String path) {
}
```

ipv6/ipv6.go
```diff
@@ -47,19 +47,34 @@ var _ = IPv6Describe("IPv6 Connectivity Tests", func() {
...
-		Expect(cf.Cf("push", appName,
-			"-m", DEFAULT_MEMORY_LIMIT,
-			"-p", buildpackPath,
-			"-s", stack,
-		).Wait(Config.DetectTimeoutDuration())).To(Exit(0))
 
-		for _, data := range EndpointTypeMap {
+		memoryLimit := DEFAULT_MEMORY_LIMIT
+		if manifestFile == "assets/java-spring/manifest.yml" {
+			memoryLimit = "1600MB" // Java Spring app specific memory limit
```
jochenehret, line 55: Couldn't this be specified in the app's `manifest.yml`?

ipv6/ipv6.go
```diff
@@ -47,19 +47,34 @@ var _ = IPv6Describe("IPv6 Connectivity Tests", func() {
...
+		}
+
+		commandOptions := []string{"push", appName, "-m", memoryLimit, "-s", stack}
+		if manifestFile != "" {
```
jochenehret, line 59: Wouldn't it be easier to add the path to the Spring app to the list of assets here:
https://github.com/cloudfoundry/cf-acceptance-tests/blob/cc7b2ce098a380008359b62a3d628681e1c19b54/helpers/assets/assets.go#L68
and then just use this folder for the `cf push -p` parameter?

ipv6/ipv6.go
```diff
@@ -47,19 +47,34 @@ var _ = IPv6Describe("IPv6 Connectivity Tests", func() {
 		},
 	}
 
-	describeIPv6Tests := func(buildpackPath string, stack string) {
+	describeIPv6Tests := func(buildpackPath, stack, manifestFile string) {
```
jochenehret, line 50: "buildpackPath" is a bit confusing here. Shouldn't it be just "asset" or "assetPath"?

jochenehret (changes requested):

ipv6/ipv6.go
```diff
@@ -47,19 +48,30 @@ var _ = IPv6Describe("IPv6 Connectivity Tests", func() {
...
-	describeIPv6Tests := func(buildpackPath string, stack string) {
+	describeIPv6Tests := func(assetPath, stack string) {
 		appName = random_name.CATSRandomName("APP")
-		Expect(cf.Cf("push", appName,
-			"-m", DEFAULT_MEMORY_LIMIT,
-			"-p", buildpackPath,
-			"-s", stack,
-		).Wait(Config.DetectTimeoutDuration())).To(Exit(0))
+
+		if assetPath == "" {
```
jochenehret, line 54: The checks for `assetPath == ""` look a bit odd. As the Java Spring app requires a different logic, can you try to move that coding into a separate function? E.g. `describeIPv6JavaSpring`?

ipv6/ipv6.go
```diff
@@ -47,25 +48,32 @@ var _ = IPv6Describe("IPv6 Connectivity Tests", func() {
...
-		).Wait(Config.DetectTimeoutDuration())).To(Exit(0))
+	pushAndValidate := func(commandOptions []string, defaultPathExpectMessage string) {
+		pushSession := cf.Cf(commandOptions...)
+		Expect(pushSession.Wait(Config.DetectTimeoutDuration())).To(Exit(0))
 
 		for _, data := range EndpointTypeMap {
 			response := helpers.CurlApp(Config, appName, data.path)
-		
+
 			if data.path == "" {
```
jochenehret, line 58: This if still looks a bit odd. Can you try:
```
for key, data := range EndpointTypeMap {
  if key == "default" {
(...)
```

jochenehret (changes requested):
Just 2 small comments, looks good overall!

assets/java-spring/target/spring-boot-trivial-app-0.0.1.jar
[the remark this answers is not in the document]
Milena-Encheva: The JAR file must be updated because the .java files have changed, adding new endpoints. Without updating the JAR file, it won't reflect these changes. This leads to errors when running the application and the tests afterwards:

```
  [FAILED] Expected
      <string>: {"timestamp":"2025-05-22T14:18:00.317+00:00","status":404,"error":"Not Found","path":"/ipv4-test"}
  to contain substring
      <string>: IPv4 validation resulted in success

```

jochenehret (approved):

PR #1573, opened 2025-05-21
IPv6 validation for Nginx

Milena-Encheva:

Yes.

The change is part of the IPv6 egress validation cycle. We progressively extend all of the exciting buildpacks relevant to our ecosystem.  In this PR we provide additional endpoints in the Nginx application for testing IPv6 support. In case ipv6 validation is enabled, we verify that egress IPv6 calls are possible with Nginx applications.  

v48.9.0

- [ ] YES
- [x] N/A

_CATs should validate IPv6 egress calls with Nginx application. We add changes regarding to this buildpack only. The test group for ipv6 was already created.

Around 90 seconds per test.

@oliver-heinrich @iaftab-alam

### DIFF
--- a/assets/nginx/nginx.conf
+++ b/assets/nginx/nginx.conf
@@ -9,5 +9,24 @@ http {
        location = / {
             return 200 "Hello NGINX!";
        }
-   }
+
+        location /ipv4-test {
+            proxy_pass https://api4.ipify.org/?format=json;
+            proxy_set_header Host api4.ipify.org;
+            proxy_ssl_server_name on;
+           }
+
+       location /ipv6-test {
+             proxy_pass https://api6.ipify.org/?format=json;
+             proxy_set_header Host api6.ipify.org;
+             proxy_ssl_server_name on;
+            }
+
+       location /dual-stack-test {
+             proxy_pass https://api64.ipify.org/?format=json;
+             proxy_set_header Host api64.ipify.org;
+             proxy_ssl_server_name on;
+            }
+
+       }
 }
--- a/ipv6/ipv6.go
+++ b/ipv6/ipv6.go
@@ -1,6 +1,7 @@
 package ipv6
 
 import (
+	"encoding/json"
 	"fmt"
 	. "github.com/cloudfoundry/cf-acceptance-tests/cats_suite_helpers"
 	"github.com/cloudfoundry/cf-acceptance-tests/helpers/app_helpers"
@@ -11,7 +12,18 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gexec"
+	"net"
 )
+
+func isIPv4(ip string) bool {
+	parsedIP := net.ParseIP(ip)
+	return parsedIP != nil && parsedIP.To4() != nil
+}
+
+func isIPv6(ip string) bool {
+	parsedIP := net.ParseIP(ip)
+	return parsedIP != nil && parsedIP.To4() == nil
+}
 
 var _ = IPv6Describe("IPv6 Connectivity Tests", func() {
 	var appName string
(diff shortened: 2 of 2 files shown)


assets/nginx/nginx.conf
```diff
@@ -9,5 +9,24 @@ http {
        location = / {
             return 200 "Hello NGINX!";
        }
-   }
+
+        location /ipv4-test {
```
jochenehret, line 13: Indentation is not correct for ipv4.

ipv6/ipv6.go
```diff
@@ -66,6 +78,52 @@ var _ = IPv6Describe("IPv6 Connectivity Tests", func() {
...
+		defaultPathExpectMessage := "Hello"
+		pushSession := cf.Cf(commandOptions...)
+		exitCode := pushSession.Wait(Config.DetectTimeoutDuration()).ExitCode()
+
+		// Log output if app push fails to diagnose deployment errors. This is crucial for
+		// identifying issues such as unsupported IPv6, which prevent the Nginx app from
+		// being successfully deployed. Capturing stdout and stderr provides insights into
+		// configuration or connectivity problems.
+
+		if exitCode != 0 {
```
jochenehret, line 92: The default pattern we have for testing a `cf push` is:
```
					Expect(cf.Cf("push",
						appName
					).Wait(Config.DetectTimeoutDuration())).To(Exit(0))
```
And then checking the app's outputs in an `AfterEach` function:
```
	AfterEach(func() {
		app_helpers.AppReport(appName)

		Expect(cf.Cf("delete", appName, "-f", "-r").Wait()).To(Exit(0))
	})
```

E.g. https://github.com/cloudfoundry/cf-acceptance-tests/blob/cd5c27ef81e7e81dfa32f99b5aa28e8b2fed5506/detect/buildpacks.go#L29

Can you apply this pattern?

ipv6/ipv6.go
```diff
@@ -11,8 +12,19 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gexec"
+	"net"
 )
 
+func isIPv4(ip string) bool {
```
jochenehret, line 18: Is this only used for the nginx test? If yes, can you try to move it into the `describeIpv6NginxTest` function?

jochenehret (changes requested):

jochenehret (dismissed):

assets/nginx/nginx.conf
```diff
@@ -9,5 +9,24 @@ http {
...
+
+       location /ipv4-test {
+            proxy_pass https://api4.ipify.org/?format=json;
+            proxy_set_header Host api4.ipify.org;
+            proxy_ssl_server_name on;
+           }
+
+       location /ipv6-test {
+             proxy_pass https://api6.ipify.org/?format=json;
+             proxy_set_header Host api6.ipify.org;
```
iaftab-alam, line 21: Not to be picky... indentation is inconsistent, for better readability and maintainability can we make it consistent?

assets/nginx/nginx.conf
```diff
@@ -9,5 +9,24 @@ http {
...
+             proxy_ssl_server_name on;
+            }
+
+       location /dual-stack-test {
+             proxy_pass https://api64.ipify.org/?format=json;
+             proxy_set_header Host api64.ipify.org;
```
iaftab-alam, line 27: The proxy_set_header and proxy_pass directives are hardcoded to a specific upstream hosts and URLs. What if we want to adapt it to other points? This is also little hard and whether it is worth todo it or not like make them configurable. But I mean if we stick to these hosts and URLs and will not change it we are good to go...

iaftab-alam (changes requested):

assets/nginx/nginx.conf
Milena-Encheva: I explored several ways to externalize or reuse the values, including defining them with set, using map, and moving them to separate files via include. However, after testing, none of these approaches worked reliably in this context.
This could be part of additional discussion, but given that these URLs are already hardcoded across all the other buildpacks, maintaining consistency by keeping them inline here seems the most straightforward and predictable option for now.

iaftab-alam (approved):

PR #1587, opened 2025-05-29
IPv6 validation for Ruby apps

Milena-Encheva:

Yes.

The change is part of the IPv6 egress validation cycle. We progressively extend all of the exciting buildpacks relevant to our ecosystem.  In this PR we provide additional endpoints in the ruby application for testing IPv6 support. In case ipv6 validation is enabled, we verify that egress IPv6 calls are possible with ruby applications.  

v48.9.0

- [ ] YES
- [x] N/A

_CATs should validate IPv6 egress calls with Ruby application. We add changes regarding to this buildpack only. The test group for ipv6 was already created.

Around 90 seconds per test.

@oliver-heinrich @iaftab-alam


philippthun:
Here are some suggestions (from Copilot) for improvement in this PR based on the diff provided:

### 1. **Error Message Handling in Sinatra Route**
In the new Ruby code, the response always includes `Error message: #{result[:error]}` even when there is no error. This will print `Error message: ` with a nil value for successful requests, which can be confusing to clients and test logs.

**Suggestion:**  
Only include the error message if there is an error.

```ruby
# Before:
"#{validation_name} validation resulted in #{result[:success] ? 'success' : 'failure'}. Detected IP type is #{result[:ip_type]}. Error message: #{result[:error]}"

# After:
message = "#{validation_name} validation resulted in #{result[:success] ? 'success' : 'failure'}. Detected IP type is #{result[:ip_type]}."
message += " Error message: #{result[:error]}" if result[:error]
message
```

---

### 2. **Potential Hardcoding & DRY Principle**
The `IPTester` class only ever receives a single endpoint, but is initialized with an array (`endpoints`). The array is not used elsewhere, so this could be simplified.

**Suggestion:**  
Refactor to remove the unused `@endpoints` unless you plan to expand this later.

---

### 3. **Error Logging Granularity**
Currently, in `test_endpoint`, any exception results in a log message with just the exception object (`#{e}`). It could be more useful to include the full backtrace for debugging.

**Suggestion:**  
Log the full backtrace for exceptions, or at least the class and message.

```ruby
logger.error("Failed to reach #{endpoint}: #{e.class} - #{e.message}\n#{e.backtrace.join("\n")}")
```

---

### 4. **IP Type Detection**
The `determine_ip_type` method returns `'IPv6'` for anything that’s not IPv4. Consider explicitly checking for both types and handling unexpected cases.

**Suggestion:**  
Return `'Unknown'` when the type is neither IPv4 nor IPv6.

```ruby
def determine_ip_type(ip_string)
  ip = IPAddr.new(ip_string)
  return 'IPv4' if ip.ipv4?
  return 'IPv6' if ip.ipv6?
  'Unknown'
rescue IPAddr::InvalidAddressError
  'Invalid IP'
end
```

---

### 5. **Test Coverage**
While the Go test adds a Ruby test case, it only checks for a `"Healthy"` response. It would be better to actually validate the IPv6 functionality and the endpoints being exercised, perhaps by hitting the new `/ipv6-test` endpoint and asserting on the result.

**Suggestion:**  
Expand the Go test to exercise the new Ruby endpoints and check for expected behavior, not just `"Healthy"`.

---

### 6. **Code Formatting**
There are some double blank lines in the Ruby file that could be reduced to keep the code tidy.

PR #1844, opened 2026-02-23
Make internal cc dns name configurable

c0d1ngm0nk3y:

Yes

Make the reference to the internal address of the cloudcontroller (`cloud-controller-ng.service.cf.internal`) configurable. This makes it possible to reuse the ASG test if you have another internal domain (e.g. `kind`)

Enable for `include_security_groups` [kind](https://github.com/cloudfoundry/kind-deployment). The default is the old hardcoded value, so no user has to change anything.

- [ ] YES
- [X] N/A

0

@nicolasbender @modulo11

### DIFF
--- a/helpers/config/config.go
+++ b/helpers/config/config.go
@@ -69,6 +69,7 @@ type CatsConfig interface {
 	GetExistingClientSecret() string
 	GetGoBuildpackName() string
 	GetHwcBuildpackName() string
+	GetInternalCCAddress() string
 	GetIsolationSegmentName() string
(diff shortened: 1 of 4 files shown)


jochenehret (commented):
Looks good, but can you also extract the hard-coded dns name here:
https://github.com/cloudfoundry/cf-acceptance-tests/blob/921575a2d23660a3ce48cd248080b5bcb746c1c8/security_groups/comma_delmited_asgs.go#L47

jochenehret (approved):

PR #1870, opened 2026-03-16
Acceptance tests for service credential binding rotations

jochenehret:

New acceptance tests for the service credential binding rotations feature. 

The first test creates one service binding for the test app / test service and records its guid. Then it creates a second binding (with the `multiple` strategy), restages the test app and ensures that the binding guid has changed. This validates the full "credential rotation" scenario.

The other tests create two service bindings for the test app / test service and then check the following actions:
1. `cf service <service name>` shows all bindings
2. `cf unbind-service <app name> <service name>` deletes all bindings
3. `cf cleanup-outdated-service-bindings` deletes the oldest binding

https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0040-service-binding-rotation.md

v54.14.0 (with `cc.max_service_credential_bindings_per_app_service_instance` set to `2`)

- [x] requires an update to a CATs integration-config (new config option `include_service_credential_binding_rotation`, default is false)

- [x] YES
- [ ] N/A

This is a new major CF feature, see [RFC-0040](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0040-service-binding-rotation.md).

Each test takes about 4 minutes to complete, so an additional of 12 minutes (if test suite is enabled).

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ include_app_syslog_tcp
 * `comma_delim_asgs_enabled`: Defaults to `false`. Set to true if comma delimited ASG destinations are enabled in the test environment.
 * `include_services`: Flag to include test for the services API.
 * `include_service_instance_sharing`: Flag to include tests for service instance sharing between spaces. `include_services` must be set for these tests to run. The `service_instance_sharing` feature flag must also be enabled for these tests to pass.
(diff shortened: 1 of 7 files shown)


oliver-heinrich (approved):
LGTM
@Milena-Encheva, can you also review and approve?

philippthun (approved):

PR #1871, opened 2026-03-18
Reduce size of catnip image

c0d1ngm0nk3y:

Yes

Reducing the size of the catnip image drastically

No need to include the complete `go` build chain on the image. Now it uses a builder to build the app and puts it on an `alpine` image.

* Size is now < 100MB instead of > 1 GB
* Multi architecture is enabled
* Dependencies (e.g. `bash`, `curl`) are now explicit and not part of `golang`.
 
v21.0.0

- [ ] YES
- [X] N/A

A bit less, depending on your internet speed and if the image already has been downloaded before.

- [X] Not urgent

@pbusko

### DIFF
--- a/assets/catnip/Dockerfile
+++ b/assets/catnip/Dockerfile
@@ -1,8 +1,15 @@
-FROM golang
+FROM --platform=$BUILDPLATFORM golang AS builder
+ARG TARGETOS
+ARG TARGETARCH
 WORKDIR /go/src/app
 COPY ./ ./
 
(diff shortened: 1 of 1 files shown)


pbusko:
@jochenehret could you please take a look at the PR?

c0d1ngm0nk3y:
I have updated the description. I thought the provided information was enough. Sorry for that. Just ping me if something is missing.

jochenehret (approved):

PR #1911, opened 2026-04-14
Implement hash based routing test

b1tamara:

Yes

Hash-based routing discussed in the [RFC0042](https://github.com/cloudfoundry/community/issues/1266) has been recently released with cf-deployment [v55.2.0](https://github.com/cloudfoundry/cf-deployment/releases/tag/v55.2.0)
The current PR extends the existing tests for the per-route options and provides implementations for a new acceptance test to ensure the hash-based routing works e2e as expected.

* [[RFC0042] Hash-Based Load Balancing Algorithm for CF Routing](https://github.com/cloudfoundry/community/issues/1266)
* [Support Hash-Based Routing in cloud_controller](https://github.com/cloudfoundry/cloud_controller_ng/pull/4746)
* [Implement hash-based routing in routing-release](https://github.com/cloudfoundry/routing-release/pull/519)

cf-deployment v55.4.0

- [X] introduces a new test --- Are you sure everyone should be running this test?

- [ ] YES
- [X] N/A

N/A

@hoffmaen

### DIFF
--- a/routing/per_route_options.go
+++ b/routing/per_route_options.go
@@ -84,6 +84,9 @@ var _ = RoutingDescribe("Per-Route Options", func() {
 		AfterEach(func() {
 			app_helpers.AppReport(appName)
 			Expect(cf.Cf("delete", appName, "-f", "-r").Wait()).To(Exit(0))
+			workflowhelpers.AsUser(TestSetup.AdminUserContext(), TestSetup.ShortTimeout(), func() {
+				Expect(cf.Cf("disable-feature-flag", "hash_based_routing").Wait()).To(Exit(0))
+			})
 		})
 
 		Context("when it's set to round-robin", func() {
--- a/routing/per_route_options.go
+++ b/routing/per_route_options.go
@@ -41,20 +41,6 @@ var _ = RoutingDescribe("Per-Route Options", func() {
 		return fmt.Sprintf("%s%s.%s", Config.Protocol(), host, Config.GetAppsDomain())
 	}
 
-	// Helper function to create busy background requests on a specific CF instance
-	createBusyInstance := func(url, instanceIndex string, count int) *sync.WaitGroup {
-		var wg sync.WaitGroup
-		for i := 0; i < count; i++ {
-			wg.Add(1)
-			go func() {
-				defer wg.Done()
-				defer GinkgoRecover()
-				helpers.Curl(Config, fmt.Sprintf("%s/delay/10", url), "-H", fmt.Sprintf("X-Cf-App-Instance: %s:%s", appId, instanceIndex))
-			}()
-		}
-		return &wg
-	}
-
 	Context("when an app sets the loadbalancing algorithm", func() {
 		BeforeEach(func() {
 			workflowhelpers.AsUser(TestSetup.AdminUserContext(), TestSetup.ShortTimeout(), func() {
@@ -156,10 +142,6 @@ var _ = RoutingDescribe("Per-Route Options", func() {
 				It("routes requests to the same instance", func() {
 					doraUrl := buildUrl(hashBasedRoutingHost)
 					hashHeader := "X-Hash-Header: 1"
-					cfInstanceIndex := helpers.Curl(Config, fmt.Sprintf("%s/env/INSTANCE_INDEX", doraUrl), "-H", hashHeader).Wait().Out.Contents()
-
-					// Create background load on the target instance
-					wg := createBusyInstance(doraUrl, string(cfInstanceIndex), 10)
 
 					reqCount := [2]int{0, 0}
 					for i := 0; i < 20; i++ {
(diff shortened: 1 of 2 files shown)


routing/per_route_options.go
```diff
@@ -123,5 +151,52 @@ var _ = RoutingDescribe("Per-Route Options", func() {
...
 		})
+		Context("when it's set to hash", func() {
+			Context("when the requests contain the same hash header", func() {
+				It("routes requests to the same instance", func() {
+					doraUrl := buildUrl(hashBasedRoutingHost)
+					hashHeader := "X-Hash-Header: 1"
+					cfInstanceIndex := helpers.Curl(Config, fmt.Sprintf("%s/env/INSTANCE_INDEX", doraUrl), "-H", hashHeader).Wait().Out.Contents()
+
+					// Create background load on the target instance
+					wg := createBusyInstance(doraUrl, string(cfInstanceIndex), 10)
```
hoffmaen: Do we need to create busy instances in this test case?

routing/per_route_options.go
```diff
@@ -123,5 +151,52 @@ var _ = RoutingDescribe("Per-Route Options", func() {
...
+					// All requests with the same hash should go to the same instance
+					Expect(reqCount[0] == 20 || reqCount[1] == 20).To(BeTrue(), "All 20 requests should be routed to the same instance")
+					wg.Wait()
+				})
+			})
+			Context("when the requests contain the different hash headers", func() {
+				It("distributes requests evenly", func() {
+					doraUrl := buildUrl(hashBasedRoutingHost)
+					// Create background load on the instance with index 0
+					wg := createBusyInstance(doraUrl, "0", 10)
```
hoffmaen: Same here, we don't need background connections in this test, do we?

routing/per_route_options.go
```diff
@@ -123,5 +151,52 @@ var _ = RoutingDescribe("Per-Route Options", func() {
...
+						uuid := make([]byte, 16)
+						rand.Read(uuid)
+						uuidStr := fmt.Sprintf("%x", uuid)
+
+						id := helpers.Curl(Config, fmt.Sprintf("%s/id", doraUrl), "-H", fmt.Sprintf("X-Hash-Header: %s", uuidStr)).Wait().Out.Contents()
+						reqCount[slices.Index(instanceIds[:], string(id))] += 1
+					}
+
+					// allow for some wiggle-room
+					expectedWithTolerance := (requestsToSent / 2) - (requestsToSent * 10 / 100)
```
hoffmaen: What's the idea behind this formula? Can you describe it in a comment? And why `*10 / 100`?

routing/per_route_options.go
b1tamara: You are right, it is not required. My goal for both tests was to have something like a base load and then curl. Do you have concerns?

routing/per_route_options.go
b1tamara: Half of the requests (with a 10% tolerance) with a random hash_header should be handled by a single instance to ensure even distribution. Does it make sense to you?

routing/per_route_options.go
hoffmaen: The base load should not have any impact on the routing, and is irrelevant for the test, right? In this case, I wouldn't create base load.

routing/per_route_options.go
hoffmaen: On the right side of the subtraction, it is `requestsToSent * 10 / 100`. `requestsToSent` is 100, so it is (100*10)/100, which is 10. Ideally, the number of requests per instance is around 50, so it's a 20% tolerance. If you were using hard-coded values for the expected number of requests (or hard-code the tolerance), it would be easier to understand IMO.

routing/per_route_options.go
```diff
@@ -123,5 +151,52 @@ var _ = RoutingDescribe("Per-Route Options", func() {
...
+				})
+			})
+			Context("when the requests contain the different hash headers", func() {
+				It("distributes requests evenly", func() {
+					doraUrl := buildUrl(hashBasedRoutingHost)
+					// Create background load on the instance with index 0
+					wg := createBusyInstance(doraUrl, "0", 10)
+
+					reqCount := [2]int{0, 0}
+					requestsToSent := 100
```
hoffmaen:
```suggestion
					requestsToSend := 100
```

routing/per_route_options.go
```diff
@@ -123,5 +151,52 @@ var _ = RoutingDescribe("Per-Route Options", func() {
...
+					for i := 0; i < requestsToSent; i++ {
+						// Generate random hash header
+						uuid := make([]byte, 16)
+						rand.Read(uuid)
+						uuidStr := fmt.Sprintf("%x", uuid)
```
hoffmaen:
```suggestion
						randomHashValue := uuid.New().String()
```

routing/per_route_options.go
```diff
@@ -1,6 +1,7 @@
 package routing
 
 import (
+	"crypto/rand"
```
hoffmaen, line 4:
```suggestion
    "github.com/google/uuid"
```

routing/per_route_options.go
hoffmaen: If you want to keep using rand, maybe use a more meaningful variable name, such as the proposed one (`randomHashValue`).

routing/per_route_options.go
b1tamara: You are right.

routing/per_route_options.go
b1tamara: I would use rand as it is used in other tests, and I do not need to update go.mod file. But I adapted the variable name. Thank you

routing/per_route_options.go
b1tamara: Removed.

routing/per_route_options.go
b1tamara: Thank you for the suggestion, but I do not want to introduce a new dependency. I will keep crypto/rand.

routing/per_route_options.go
b1tamara: Removed

routing/per_route_options.go
```diff
@@ -123,5 +137,44 @@ var _ = RoutingDescribe("Per-Route Options", func() {
...
+						uuid := make([]byte, 16)
+						rand.Read(uuid)
+						randomHashValue := fmt.Sprintf("%x", uuid)
+
+						id := helpers.Curl(Config, fmt.Sprintf("%s/id", doraUrl), "-H", fmt.Sprintf("X-Hash-Header: %s", randomHashValue)).Wait().Out.Contents()
+						reqCount[slices.Index(instanceIds[:], string(id))] += 1
+					}
+
+					// allow for some wiggle-room
+					tolerance := 5
```
hoffmaen: Are we confident that the probability to be inside the tolerance is high enough? I'd feel better with a larger value for tolerance.

b1tamara: The tests are passed with 5. But you are right, it might be flacky. Changed to 10.

hoffmaen (approved):

jochenehret:
So this new test requires cf-deployment >= v55.2.0, correct? Then we should cut a new CATs release with according release notes.

b1tamara:
@jochenehret, yes.
Hash-based routing has been released with [capi-release 1.225](https://github.com/cloudfoundry/capi-release/releases/tag/1.225.0) and [routing-release v0.374.0](https://github.com/cloudfoundry/routing-release/releases/tag/v0.374.0). The routing release v.0.374.0 is part of [v55.2.0](https://github.com/cloudfoundry/cf-deployment/releases/tag/v55.2.0)

jochenehret (approved):
Tested with capi-release 1.231.0 and routing-release 0.374.0, works.

PR #1984, opened 2026-06-09
Build dotnet-core asset for cflinuxfs5 (net8.0/linux-x64)

Milena-Encheva:

Adds the compiled cflinuxfs5 binary for the dotnet-core test asset, built targeting net8.0 with the linux-x64 runtime identifier. Updates the **Makefile** to use linux-x64 instead of ubuntu.24.04-x64 for the cflinuxfs5 build target, as ubuntu.24.04-x64 is not recognized by the .NET 6.0 SDK when MSBuild evaluates the multi-target project, causing the build to fail.

The **dotnet-core.csproj** targets both net6.0 and net8.0. When publishing with -f net8.0 -r ubuntu.24.04-x64, MSBuild evaluates all target frameworks including net6.0, which does not support the ubuntu.24.04-x64 RID, resulting in error NETSDK1083. Switching to the portable linux-x64 RID is compatible with both framework versions and produces an equivalent self-contained binary for cflinuxfs5 (Ubuntu 24.04).

N/A — this change only updates the pre-compiled test asset binary and its [Makefile](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) build target. No test logic was added or modified.

- [ ] YES
- [X] N/A

N/A — no new test is being introduced.

0 — no test logic was changed. This only adds a missing compiled asset needed to run existing dotnet-core tests against the cflinuxfs5 stack.

- [X] **Urgent** - unblocks running CATs against cflinuxfs5 stack

@oliver-heinrich @jochenehret


assets/dotnet-core/cflinuxfs5/cflinuxfs4/cflinuxfs3/appsettings.Development.json
jochenehret, line 1: Is this nested file structure intended? I would say we could remove cflinuxfs3 remainders.

Milena-Encheva: The nested folders are generated by running make in the dotnet-core folder. 
The cflinuxfs3 paths you see are generated artifact remnants inside publish outputs.
I can open a follow-up cleanup PR to remove cflinuxfs3 leftovers.

jochenehret (approved):

PR #2033, opened 2026-07-15
Add syslog filter test

jorbaum:

A new feature has been added to the syslog-agent of the loggregator-agent-release. This adds some initial tests to make sure users can use the feature as expected.

PR: https://github.com/cloudfoundry/loggregator-agent-release/pull/714

Issue: https://github.com/cloudfoundry/loggregator-agent-release/issues/711

Documentation PRs:

* https://github.com/cloudfoundry/docs-dev-guide/pull/576
* https://github.com/cloudfoundry/docs-dev-guide/pull/609
* https://github.com/cloudfoundry/docs-dev-guide/pull/608

I deployed CF-on-K8s using cloudfoundry/kind-deployment at commit [db520282](https://github.com/cloudfoundry/kind-deployment/commit/db5202825b28baa2d36bc6b4899b152cc58d13a4) (v0.22.0).

- [ ] YES
- [x] N/A

This is a new feature for syslog drains.

On my machine tests run through (including skipping tests) in roughly 8 minutes:

```bash
CONFIG="$HOME/Workspace/public/cloudfoundry/kind-deployment/.github/cats-config.json" go run github.com/onsi/ginkgo/v2/ginkgo --no-color --timeout=45m \
  --focus="Syslog Drain source type filter over TCP" .
  ...
Ran 5 of 291 Specs in 1132.745 seconds
SUCCESS! -- 5 Passed | 0 Failed | 2 Pending | 284 Skipped
```

@beyhan

**Disclaimer**: Claude Opus 4.8 (and Claude Code) generated this code, I reviewed carefully and edited.

### DIFF
--- a/app_syslog_tcp/syslog_drain_filter.go
+++ b/app_syslog_tcp/syslog_drain_filter.go
@@ -1,0 +1,141 @@
+package apps
+
+import (
+	"fmt"
+	"time"
+
+	. "github.com/cloudfoundry/cf-acceptance-tests/cats_suite_helpers"
+	"github.com/cloudfoundry/cf-acceptance-tests/tcp_routing"
+
+	"github.com/cloudfoundry/cf-acceptance-tests/helpers/app_helpers"
+	"github.com/cloudfoundry/cf-acceptance-tests/helpers/assets"
+	logshelper "github.com/cloudfoundry/cf-acceptance-tests/helpers/logs"
+	"github.com/cloudfoundry/cf-acceptance-tests/helpers/random_name"
+	"github.com/cloudfoundry/cf-test-helpers/v2/cf"
+	"github.com/cloudfoundry/cf-test-helpers/v2/helpers"
+	"github.com/cloudfoundry/cf-test-helpers/v2/workflowhelpers"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+)
+
+// Verifies the per-drain log source type filter from the syslog-agent. HTTP traffic
+// against a ruby_simple app produces both APP logs (the appMarker echoed to STDOUT)
+// and RTR logs (gorouter access logs).
+const rtrSourceTypeMarker = `source_type="RTR"`
+
+var _ = AppSyslogTcpDescribe("Syslog Drain source type filter over TCP", func() {
+	var (
+		logWriterAppName string
+		externalPort     string
+		domainName       string
+		listenerAppName  string
+		logs             *Session
+		interrupt        chan struct{}
+		serviceName      string
+	)
+
+	Describe("Syslog drain source type filter", func() {
+		BeforeEach(func() {
+			interrupt = make(chan struct{}, 1)
+			domainName = Config.GetTCPDomain()
+			workflowhelpers.AsUser(TestSetup.AdminUserContext(), Config.DefaultTimeoutDuration(), func() {
+				routerGroupOutput := string(cf.Cf("router-groups").Wait().Out.Contents())
+				Expect(routerGroupOutput).To(
+					MatchRegexp(fmt.Sprintf("%s\\s+tcp", tcp_routing.DefaultRouterGroupName)),
+					fmt.Sprintf("Router group %s of type tcp doesn't exist", tcp_routing.DefaultRouterGroupName),
+				)
+
+				Expect(cf.Cf("create-shared-domain",
+					domainName,
+					"--router-group", tcp_routing.DefaultRouterGroupName,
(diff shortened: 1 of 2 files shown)


chombium (changes requested):
Hi @jorbaum,

the query parameters should be adjusted as they were changed in the Syslog Agents code from `include-log-source-types` and `exclude-log-source-types` to `include-log-types` and `exclude-log-types`.

A part form that LGTM.

chombium (approved):
LGTM! Thanks @jorbaum

app_syslog_tcp/syslog_drain_filter.go
```diff
@@ -0,0 +1,186 @@
...
+			interrupt = make(chan struct{}, 1)
+			domainName = Config.GetTCPDomain()
+			workflowhelpers.AsUser(TestSetup.AdminUserContext(), Config.DefaultTimeoutDuration(), func() {
+				routerGroupOutput := string(cf.Cf("router-groups").Wait().Out.Contents())
+				Expect(routerGroupOutput).To(
+					MatchRegexp(fmt.Sprintf("%s\\s+tcp", tcp_routing.DefaultRouterGroupName)),
+					fmt.Sprintf("Router group %s of type tcp doesn't exist", tcp_routing.DefaultRouterGroupName),
+				)
+
+				Expect(cf.Cf("create-shared-domain",
```
jochenehret, line 49: This shared domain should be deleted in the test's teardown.

app_syslog_tcp/syslog_drain_filter.go
```diff
@@ -0,0 +1,186 @@
...
+			app_helpers.AppReport(listenerAppName)
+
+			Eventually(cf.Cf("delete", logWriterAppName, "-f", "-r")).Should(Exit(0), "Failed to delete log writer app")
+			Eventually(cf.Cf("delete", listenerAppName, "-f", "-r")).Should(Exit(0), "Failed to delete listener app")
+			Eventually(cf.Cf("delete-service", serviceName, "-f")).Should(Exit(0), "Failed to delete service")
+
+			Eventually(cf.Cf("delete-orphaned-routes", "-f"), Config.CfPushTimeoutDuration()).Should(Exit(0), "Failed to delete orphaned routes")
+		})
+
+		It("include-log-types=APP forwards APP logs and drops RTR logs", func() {
```
jochenehret: The body of the `It()` blocks is almost identical. Can you extract it into a reusable function?
```
assertDrainBehavior := func(drainURL, scenario string, expectRTR bool) {
...
}
```

jochenehret (changes requested):
Looks good overall, but I'm a bit concerned about the additional time these tests take.
Would it be possible to push the test apps only once for all tests? (using `BeforeAll()` instead of `BeforeEach()`) As I understand, the test apps don't have any state. Only the service should be deleted and recreated per test.

app_syslog_tcp/syslog_drain_filter.go
jorbaum: Yes. Good find.

jorbaum:
> Looks good overall, but I'm a bit concerned about the additional time these tests take. Would it be possible to push the test apps only once for all tests? (using `BeforeAll()` instead of `BeforeEach()`) As I understand, the test apps don't have any state. Only the service should be deleted and recreated per test.

Thanks for the review. Your comments are all valid and make the code better.

Now it takes 9 minutes instead of 18 on my machine/kind:

```
$ scripts/cats.sh --focus="Syslog Drain over TCP" --timeout=45m (in cloudfoundry/kind-deployment)
...
Ran 5 of 295 Specs in 499.901 seconds
SUCCESS! -- 5 Passed | 0 Failed | 2 Pending | 288 Skipped
```

I could also reduce the amount of tests to reduce the time even further if that helps.

jorbaum:
I also harmonized `app_syslog_tcp/syslog_drain.go` with the newest changes of this branch.

app_syslog_tcp/syslog_drain_filter.go
jochenehret: Ah sorry I was wrong: The shared TCP domain is used by multiple tests. Deleting it can cause race conditions. So just remove the "delete-shared-domain" commands from this PR. We'll check how to clean up domains in another PR.

jochenehret (changes requested):
Works fine, but we must avoid deleting the shared domain to avoid race conditions.

app_syslog_tcp/syslog_drain_filter.go
jorbaum: Reverted the change by dropping some commits. Now my new run takes 11 minutes with 2 parallel processes instead of 4 (changed due to RAM issues on my machine).

jochenehret (approved):

PR #2104, opened 2026-09-08
Add CATS tests for async recursive delete behavior

WeiQuan0605:

Yes

It adds acceptance tests that verify CF correctly handles async service unbind when deleting a service instance or app:

  - cf delete-service <instance> -f — completes successfully even when the broker returns 202 for unbind
  - cf delete <app> -f — completes successfully even when the bound service broker returns 202 for unbind
  - A baseline test for the synchronous unbind path (broker returns 200)

  Also wires up the include_async_recursive_delete config flag across the config system, README, and skip messages so the tests can be opted into per environment.

issue: https://github.com/cloudfoundry/cloud_controller_ng/issues/3589

cf-deployment v60.5.0

- [X] YES
- [ ] N/A

Async service unbind during app/service instance deletion is a common operator workflow — operators regularly delete apps or service instances that have active service bindings. This test validates that CF correctly handles the case where the service broker requires async unbind, which is a cross-component behavior involving CAPI, the CC worker, and the service broker protocol. It belongs in CATs because it exercises the interaction between these components in a real CF deployment.

 Each test does: push app + create service + async bind + async unbind/delete + polling. Roughly:
  - Test 1 (cf delete-service): ~3-5 minutes
  - Test 2 (cf delete): ~3-5 minutes
  - Test 3 (sync unbind baseline): ~2-3 minutes

  Total: approximately 8-13 minutes of additional runtime when `include_async_recursive_delete: true`. Since it defaults to false, there is zero runtime impact for environments that don't opt in.

### DIFF
--- a/cats_suite_helpers/cats_suite_helpers.go
+++ b/cats_suite_helpers/cats_suite_helpers.go
@@ -283,6 +283,20 @@ func ServicesDescribe(description string, callback func()) bool {
 	})
 }
 
+func AsyncRecursiveDeleteDescribe(description string, callback func()) bool {
+	return Describe("[async_recursive_delete]", func() {
+		BeforeEach(func() {
+			if !Config.GetIncludeServices() {
+				Skip(skip_messages.SkipServicesMessage)
+			}
+			if !Config.GetIncludeAsyncRecursiveDelete() {
+				Skip(skip_messages.SkipAsyncRecursiveDeleteMessage)
+			}
+		})
+		Describe(description, callback)
+	})
+}
+
 func ServiceInstanceSharingDescribe(description string, callback func()) bool {
 	return Describe("[service instance sharing]", func() {
 		BeforeEach(func() {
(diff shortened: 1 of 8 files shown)


cats_suite_helpers/cats_suite_helpers.go
```diff
@@ -283,6 +283,20 @@ func ServicesDescribe(description string, callback func()) bool {
 	})
 }
 
+func AsyncRecursiveDeleteDescribe(description string, callback func()) bool {
+	return Describe("[async_recursive_delete]", func() {
+		BeforeEach(func() {
+			if !Config.GetIncludeServices() {
```
jochenehret: Why is `include_services` required for this test suite?

jochenehret (commented):
Works fine, tested on capi-release 1.244.0. Only question is why `include_services` is required for this test suite.

jochenehret (approved):

